PR #2, opened 2023-06-22
Bootstrap building

ncdc:


Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM registry.ci.openshift.org/ocp/builder:rhel-8-golang-1.20-openshift-4.14 AS builder
+WORKDIR /build
+COPY . .
+RUN mkdir -p bin && \
+    go build -o bin ./cmd/cluster-olm-operator
```
joelanford, line 5: Any reason not to use `make build` here?

grokspawn (approved):

Dockerfile
ncdc: No, I added make build after I wrote this 👍

joelanford (approved):

PR #8, opened 2023-06-26
Add initial CVO manifests

ncdc:


joelanford:
I can't remember where we landed. Are we waiting to add the `release.openshift.io/feature-set: TechPreviewNoUpgrade` annotation so that we can avoid the eventual TPNU conflict with platform operators and this repo fighting over rukpak?

Or are we going ahead with TPNU here and removing rukpak from PO?

ncdc:
Oh, yeah, I need to add them. What was our ordered list of operations for doing this and testing it?

assets/catalogd/02-customresourcedefinition-catalogmetadata.catalogd.operatorframework.io.yml
joelanford, line 1: This is an interesting one. This CRD is unused unless we set `--feature-gates=CatalogMetadataAPI=true` (which we are not doing).

I think it makes sense to go ahead and merge it to reduce complexity/moving parts, but post-4.14, it might make sense to figure out how to dynamically include manifests based on enabled feature gates (similar to what CVO does)

PR #18, opened 2023-06-29
Add rukpak

dtfranz:
Adds rukpak static and deployment controllers, as well as the manifests.

Co-authored by: @ncdc @joelanford


manifests/0000_51_olm_06_deployment.yaml
```diff
@@ -45,6 +45,8 @@ spec:
           value: quay.io/openshift/origin-olm-catalogd:latest
         - name: OPERATOR_CONTROLLER_IMAGE
           value: quay.io/openshift/origin-olm-operator-controller:latest
+        - name: RUKPAK_IMAGE
+          value: quay.io/openshift/origin-olm-rukpak:latest
```
joelanford, line 49: We also need to add an image reference for this image in this file: https://github.com/openshift/cluster-olm-operator/blob/main/manifests/image-references.pending

joelanford:
We also need to add the names of all clusterroles and roles that are in the rukpak manifests to this list: https://github.com/openshift/cluster-olm-operator/blob/dfc350ab1de570bd0d02e29dea4ed91e4eb98477/manifests/0000_51_olm_02_operator_clusterrole.yaml#L161-L169

manifests/0000_51_olm_06_deployment.yaml
ncdc: I have that in my PR #16. We can discuss the order and split up

manifests/0000_51_olm_02_operator_clusterrole.yaml
```diff
@@ -167,3 +167,8 @@ rules:
     - operator-controller-metrics-reader
     - operator-controller-proxy-role
     - operator-controller-leader-election-role
+    - bundle-reader
+    - bundle-uploader
+    - core-admin
+    - helm-provisioner-admin
```
joelanford, line 173: Fine for now, but in upstream, we should make sure these (and all rukpak install objects) are all prefixed with `rukpak-` IMO.

PR #25, opened 2023-07-13
Dynamically generate controllers based on contents of assets

joelanford:


cmd/cluster-olm-operator/main.go
```diff
@@ -369,3 +182,104 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
...
+				Resource:  restMapping.Resource.Resource,
+				Namespace: manifest.GetNamespace(),
+				Name:      manifest.GetName(),
+			})
+
+			if manifest.GetKind() == "Deployment" {
+				controllerName := fmt.Sprintf("%sDeployment%s",
+					namePrefix,
+					strings.ReplaceAll(titler.String(manifest.GetName()), "-", ""),
+				)
```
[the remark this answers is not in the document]
ncdc: I'm tempted to say we do something that works for 4.14 TP, then figure out beyond how to handle renames and/or drops?

cmd/cluster-olm-operator/main.go
```diff
@@ -369,3 +182,104 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
...
+				return nil
+			}
+			relatedObjects = append(relatedObjects, configv1.ObjectReference{
+				Group:     restMapping.GroupVersionKind.Group,
+				Resource:  restMapping.Resource.Resource,
+				Namespace: manifest.GetNamespace(),
+				Name:      manifest.GetName(),
+			})
+
+			if manifest.GetKind() == "Deployment" {
```
ncdc, line 240: I'd make sure it's `apps` too

cmd/cluster-olm-operator/main.go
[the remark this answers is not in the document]
joelanford: Fine with me. I'll create an issue to keep track of this.

cmd/cluster-olm-operator/main.go
```diff
@@ -369,3 +182,105 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
 		return []byte(newDeployment), nil
 	}
 }
+
+type controllerBuilder struct {
```
ncdc, line 186: Doesn't necessarily need to be in this PR, but I think it's time to split this out into a separate file and possibly package.

cmd/cluster-olm-operator/main.go
```diff
@@ -369,3 +182,105 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
...
+	assetsFS fs.FS
+	cl       *clients.Clients
```
ncdc, line 188:
```suggestion
	clients       *clients.Clients
```

cmd/cluster-olm-operator/main.go
```diff
@@ -369,3 +182,105 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
...
+	cc       *controllercmd.ControllerContext
```
ncdc, line 189: I'd rename cc to something like ctrlCtx or controllerCtx or controllerContext

cmd/cluster-olm-operator/main.go
```diff
@@ -369,3 +182,105 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
...
+			}
+			relatedObjects = append(relatedObjects, configv1.ObjectReference{
+				Group:     restMapping.GroupVersionKind.Group,
+				Resource:  restMapping.Resource.Resource,
+				Namespace: manifest.GetNamespace(),
+				Name:      manifest.GetName(),
+			})
+
+			if manifestGVK.Kind == "Deployment" && manifestGVK.Group == "apps" {
+				controllerName := fmt.Sprintf("%sDeployment%s",
```
ncdc, line 242: This probably should be a separate function with tests

ncdc: And/or think about how we could write this so it's more testable?

ncdc:
Review comments will be in future PR.
/lgtm

PR #26, opened 2023-07-13
Switch to getting operand manifests using init containers

ncdc:


ncdc:
It looks like this change worked despite the e2e job failing. The failure appears unrelated and I believe cluster-olm-operator worked with the init containers to copy the manifests.

/retest

dtfranz (approved):
/lgtm

PR #28, opened 2023-07-14
Move controller code into its own package

joelanford:
Follows up on comments from #25

### DIFF
--- a/pkg/controller/builder_test.go
+++ b/pkg/controller/builder_test.go
@@ -1,0 +1,9 @@
+package controller
+
+import (
+	"testing"
+)
+
+func TestControllerNameForObject(t *testing.T) {
+	// TODO: Add test cases.
+}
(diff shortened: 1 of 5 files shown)


pkg/controller/builder_test.go
```diff
@@ -0,0 +1,9 @@
+package controller
+
+import (
+	"testing"
+)
+
+func TestControllerNameForObject(t *testing.T) {
+	// TODO: Add test cases.
+}
```
joelanford, line 9: Marked the PR as draft, mainly because I wanted to check to see if this is what you had in mind in terms of the scope of the testing in this comment: https://github.com/openshift/cluster-olm-operator/pull/25#discussion_r1263808519

Was it [this the function](https://github.com/openshift/cluster-olm-operator/pull/28/files#diff-a3a53f8da96ba212631a3dcbaefe688da5491bce9e55daaaf83c20b762bcac4eR131) you were thinking about testing for?

ncdc: Yes, thanks! And more over time too

PR #48, opened 2024-04-05
OCPBUGS-30491: Update kube dep to 1.27.12

kevinrizza:
Update kube to latest 1.27 to resolve [CVE-2024-24786](https://access.redhat.com/security/cve/CVE-2024-24786)


grokspawn:
I forget that we have this downstream-only piece that coordinates v1 as a third-order operator.  I don't want dependabot running on ocp payload repos per se, but I'm a little concerned if cves are the primary motive force for bumping dependencies. 

/lgtm

kevinrizza:
> I'm a little concerned if cves are the primary motive force for bumping dependencies.

Agreed, we need to have a better feedback loop on this component (and a couple more, to be honest)

grokspawn:
> > I'm a little concerned if cves are the primary motive force for bumping dependencies.
> 
> Agreed, we need to have a better feedback loop on this component (and a couple more, to be honest)

*cough* *cough* marketplace

PR #51, opened 2024-04-30
OPRUN-3291: Update dependencies

m1kola:


joelanford:
Looks like CRDs have moved around in the `openshift/api` repo. Going to ask to see if we're supposed to continue including them in our CVO manifests.

PR #53, opened 2024-05-01
OPRUN-3291: Add roles for clusterextensions

m1kola:
These are required for validatingadmissionpolicies and validatingadmissionpolicybindings to work.


joelanford:
/lgtm
/approve

/hold
Just want to see the e2e pass. I'm trying to remember if there's a reason that test isn't required. I can't think of a good reason we shouldn't make it required.

m1kola:
E2E jobs is green. Unholding

/unhold

PR #64, opened 2024-08-06
OPRUN-3368: block cluster upgrade when installed operators have a max ocp version set that is incompatible with the next  minor release

gallettilance:
DEMO

https://github.com/user-attachments/assets/2efca0e6-6e66-403c-82f3-72a0a4500229


pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+)
+
+const (
+	incompatibleOperatorsInstalled = "IncompatibleOperatorsInstalled"
+	MaxOpenShiftVersionProperty    = "olm.maxOpenShiftVersion"
+	OwnerKindKey                   = "olm.operatorframework.io/owner-kind"
+	OwnerNameKey                   = "olm.operatorframework.io/owner-name"
+	PackageNameKey                 = "olm.operatorframework.io/package-name"
+	BundleNameKey                  = "olm.operatorframework.io/bundle-name"
+	BundleVersionKey               = "olm.operatorframework.io/bundle-version"
```
joelanford, line 41: Probably makes sense to define these in the o-c API.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+	} `yaml:"chart"`
+}
+
+func NewDynamicUpgradeableConditionController(kubeclient kubernetes.Interface, client dynamic.Interface, name string, operatorClient *clients.OperatorClient, eventRecorder events.Recorder, prefixes []string) factory.Controller {
+	infFact := dynamicinformer.NewDynamicSharedInformerFactory(client, 0)
+
+	clusterExtensionGVR := schema.GroupVersionResource{
+		Group:    ocv1alpha1.ClusterExtensionGVK.Group,
+		Version:  ocv1alpha1.ClusterExtensionGVK.Version,
+		Resource: ocv1alpha1.ClusterExtensionGVK.Kind,
```
joelanford, line 71: Resource != Kind, so I don't expect this to work. 

In this case:
- `Resource` = `clusterextensions`
- `Kind` = `ClusterExtension`

We can do either of these, IMO:
1. Just hardcode `clusterextensions`
2. Use a restmapper to translate from kind to resource.

I think (1) is probably fine because changing from `clusterextensions` to anything else would break all versions of the CRD. So it is extremely unlikely to happen.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+	if !ok {
+		log.Println("Error: could not cast to ClusterExtension object")
+		return
+	}
+
+	c.syncIncompatibleOperators()
+}
+
+func (c *dynamicUpgradeableConditionController) syncIncompatibleOperators() {
+	current, err := getCurrentRelease()
```
joelanford, line 105:
```suggestion
currentOCPVersion, err := getCurrentOpenShiftVersion()
```

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+		MaxReadChunks:  0,
+		MaxWriteChunks: 0,
+		Log:            nil,
+	}
+
+	owner := ce.Name
+	ownerRefs := []metav1.OwnerReference{*metav1.NewControllerRef(ce, ce.GetObjectKind().GroupVersionKind())}
+	ownerRefSecretClient := helmclient.NewOwnerRefSecretClient(secretClient, ownerRefs, func(secret *corev1.Secret) bool {
+		return secret.Type == storage.SecretTypeChunkedIndex
+	})
```
joelanford, line 213: I don't think you need this part. If I remember correctly, this is only used to write release secrets. We're only reading from storage, so none of this ends up being used.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+	client         dynamic.Interface
+	name           string
+	operatorClient *clients.OperatorClient
+	prefixes       []string
+}
+
+type ChartMetadata struct {
+	Metadata struct {
+		Annotations map[string]string `yaml:"annotations"`
+	} `yaml:"metadata"`
```
everettraven, line 56: Is there a way this could reasonably be pulled out into a separate type? This notation threw me off at first glance.

Maybe something like:
```go
type Metadata struct {
    Annotations map[string]string `yaml:"annotations"`
}

type Chart struct {
    Metadata Metadata `yaml:"metadata"`
}
```

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+}
+
+type SecretData struct {
+	Chart struct {
+		ChartMetadata
+	} `yaml:"chart"`
```
everettraven, line 62: Is this any different than:
```go
type SecretData struct {
    Chart ChartMetadata `yaml:"chart"`
}
```
?

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+		prefixes:       prefixes,
+	}
+
+	inf.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.handleClusterExtension,
+		UpdateFunc: func(oldObj, newObj interface{}) { c.handleClusterExtension(newObj) },
+		DeleteFunc: c.handleClusterExtension,
+	})
+
+	return factory.New().WithInformers(inf.Informer()).ToController(name, eventRecorder)
```
everettraven, line 91: Looking into the `github.com/openshift/library-go/pkg/controllers/factory` package, it looks like using `WithInformers()` is for setting informers that you want to trigger the `Sync` function for this controller [1]. If you want to use custom event handlers, you probably want to use `WithBareInformers()` [2].

[1]: https://github.com/openshift/library-go/blob/7c5c5d305ac1bb911c5a4a0be82fa7d5aa7c5e35/pkg/controller/factory/factory.go#L95-L101
[2]: https://github.com/openshift/library-go/blob/7c5c5d305ac1bb911c5a4a0be82fa7d5aa7c5e35/pkg/controller/factory/factory.go#L115-L123

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+			If the version is already set, we don't want to set it again as the currentRelease
+			is designed to be a singleton. If a new version is set, we are making an assumption
+			that this controller will be restarted and thus pull in the new version from the
+			environment into memory.
+
+			Note: sync.Once is not used here as it was difficult to reliably test without hitting
+			race conditions.
+		*/
+		return currentRelease.version, nil
+	}
```
joelanford, line 250: This global variable and mutex seems like unnecessary complexity to me.
1. It protects an envvar lookup/semver parse call, so it doesn't seem like there is a performance reason to cache that result.
2. The envvar can't change unless:
    - We change it ourselves (we won't)
    - The process restarts (no different than what is already happening)

I'd suggest changing this function to always lookup the envvar and parse it on every call.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+				if err != nil {
+					log.Printf("Error converting to semver for version %s: %v", version, err)
+					continue
+				}
+				if maxOCPVersion == nil || maxOCPVersion.GTE(next) {
+					// All good
+				} else {
+					// Incompatible
+					name := rel.Labels[BundleNameKey]
+					incompatibleOperators = append(incompatibleOperators, name)
```
joelanford, line 160: I know this would differ from OLMv0, but it might be nice to include the name of the ClusterExtension somehow in the error message, which would give users a breadcrumb to the object they need to change in order to resolve this condition.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+	updateStatusFuncs := make([]v1helpers.UpdateStatusFunc, 0, 1)
+	if len(incompatibleOperators) > 0 {
+		updateStatusFuncs = append(updateStatusFuncs, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:    "Upgradeable",
+			Status:  operatorv1.ConditionFalse,
+			Reason:  incompatibleOperatorsInstalled,
+			Message: strings.Join(incompatibleOperators, ","),
+		}))
+
+		if _, _, updateErr := v1helpers.UpdateStatus(context.TODO(), c.operatorClient, updateStatusFuncs...); updateErr != nil {
```
joelanford, line 175: Do you know if this is a no-op if the set of incompatible operators has not changed?

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+		kubeclient:     kubeclient,
+		name:           name,
+		operatorClient: operatorClient,
+		prefixes:       prefixes,
+	}
+
+	inf.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.handleClusterExtension,
+		UpdateFunc: func(oldObj, newObj interface{}) { c.handleClusterExtension(newObj) },
+		DeleteFunc: c.handleClusterExtension,
```
everettraven, line 88: After taking a further look into the openshift/library-go package, we might be better off using the factory to create a controller with a specific sync function similar to what is done in https://github.com/openshift/cluster-olm-operator/blob/e31b7775ef903ff0c35f95656af27afa2dddd3e4/pkg/controller/staticupgradeableconditioncontroller.go#L23

is there a particular reason you went with doing custom event handlers?

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+	})
+
+	return factory.New().WithInformers(inf.Informer()).ToController(name, eventRecorder)
+}
+
+func (c *dynamicUpgradeableConditionController) handleClusterExtension(obj interface{}) {
+	_, ok := obj.(*ocv1alpha1.ClusterExtension)
+	if !ok {
+		log.Println("Error: could not cast to ClusterExtension object")
```
everettraven, line 97: It looks like we use `klog` in the other controller we have:
https://github.com/openshift/cluster-olm-operator/blob/e31b7775ef903ff0c35f95656af27afa2dddd3e4/pkg/controller/staticupgradeableconditioncontroller.go#L33

We should be consistent and use it here as well

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+func toSemver(max string) (*semver.Version, error) {
+	value := strings.Trim(max, "\"")
+	if value == "" {
+		// Handle "" separately, so parse doesn't treat it as a zero
+		return nil, fmt.Errorf(`value cannot be "" (an empty string)`)
+	}
+
+	version, err := semver.ParseTolerant(value)
+	if err != nil {
+		return nil, fmt.Errorf(`failed to parse "%s" as semver: %w`, value, err)
```
joelanford, line 292: Nit:
```suggestion
return nil, fmt.Errorf(`failed to parse %q as semver: %w`, value, err)
```

pkg/controller/dynamicupgradeableconditionscontroller.go
joelanford: This notation is somewhat common when declaring one-off structs for parsing like we are in this case. I actually prefer the current PR's setup in that regard.

Though I do notice that `ChartMetadata` is exported. We should unexport that.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+	} `yaml:"metadata"`
+}
+
+type SecretData struct {
+	Chart struct {
+		ChartMetadata
+	} `yaml:"chart"`
+}
+
+func NewDynamicUpgradeableConditionController(kubeclient kubernetes.Interface, client dynamic.Interface, name string, operatorClient *clients.OperatorClient, eventRecorder events.Recorder, prefixes []string) factory.Controller {
```
joelanford, line 65: I didn't check all of the function parameters, but are there any we don't use? I see we set `prefixes` on the unexported struct, but I don't actually see it used anywhere.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,300 @@
...
+	infFact := dynamicinformer.NewDynamicSharedInformerFactory(client, 0)
```
joelanford, line 66: What does it mean to pass `0` as the resync period here? I wonder if we should set that to something like `10 * time.Hour`, which I think what controller-runtime uses as a default.

pkg/controller/dynamicupgradeableconditionscontroller.go
gallettilance: I want this to be triggered on clusterextension events. If there are no changes with clusterextensions, there are no changes to the set of incompatible operators.

pkg/controller/dynamicupgradeableconditionscontroller.go
gallettilance: agreed - let's create a new ticket to track that addition and I can create a follow up PR

pkg/controller/dynamicupgradeableconditionscontroller.go
everettraven: I think you can still achieve that using a similar structure as above. It's now been a week or so since I've looked at it though so I'd have to look again to verify.

cmd/cluster-olm-operator/main.go
```diff
@@ -119,7 +128,7 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 
 	cl.StartInformers(ctx)
 
-	for _, c := range append(controllerList, upgradeableConditionController, clusterOperatorController) {
+	for _, c := range append(controllerList, staticUpgradeableConditionController, dynamicUpgradeableConditionController, clusterOperatorController) {
```
joelanford, line 131: I haven't dived in to look further, but it seems like it could be a problem to have two different controllers trying to control the upgradeable condition, right?

joelanford: Looked a little further the `statusUpgradeableConditionController` only seems to set `${prefix}Upgradeable` conditions (good!), but then what (if anything) is aggregating those up to a single `Upgradeable` condition?

pkg/controller/dynamicupgradeableconditionscontroller.go
gallettilance: I don't think there's an interface where I can omit this. In order for the helm.Storage to be defined I need a driver defined which requires all this jazz

cmd/cluster-olm-operator/main.go
gallettilance: yeah I thought the update function would only update prefix specific types but after some testing this afternoon I realized it wasn't the case - moved that sync in the dynamicUpgradeableConditionController code

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,337 @@
...
+func nextY(v semver.Version) (semver.Version, error) {
+	v.Build = nil // Builds are irrelevant
+
+	if len(v.Pre) > 0 {
+		// Dropping pre-releases is equivalent to incrementing Y
+		v.Pre = nil
+		v.Patch = 0
+
+		return v, nil
+	}
```
joelanford, line 315: Based on the conversation we had with @wking, should we drop this logic that causes pre-release versions to be treated the same as previous minors?

pkg/controller/dynamicupgradeableconditionscontroller.go
LalatenduMohanty, line 1: Kind of an cosmetic issue, but IMO the name is too long for a controller or the as a file name.

LalatenduMohanty: This is also a nitpick, we can fix this after this PR is merged.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,251 @@
...
+
+	ceList, err := c.clusterExtensionClient.Informer().Lister().List(labels.NewSelector())
+	if err != nil {
+		c.logger.V(4).Error(err, "Error listing cluster extensions")
+		return nil, err
+	}
+
+	// Get all ClusterExtensions incompatible with next Y-stream
+	for _, obj := range ceList {
+		ce, ok := obj.(*ocv1alpha1.ClusterExtension)
```
perdasilva, line 138: Minior suggestion: this seems to be essentially a filter function. It might be worth to define a type like ExtensionCompatibilityFunc or something, then give this class/struct a list of these functions. This would help make the design additive and then the next time we want to add a different check, just implement the function, add it to the list, and bob's your uncle (assuming all this function would need is the CE and the corresponding release info).

pkg/controller/dynamicupgradeableconditionscontroller.go
gallettilance: will create a PR

pkg/controller/dynamicupgradeableconditionscontroller.go
gallettilance: how about just removing the "dynamic" part of the name?

LalatenduMohanty: Just `upgradeablecontroller.go` should be fine IMO

pkg/controller/dynamicupgradeableconditionscontroller.go
LalatenduMohanty: @gallettilance I will recommend to create an upstream issue (in operator-controller) before creating a PR.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+
+	updateStatusFuncs := make([]v1helpers.UpdateStatusFunc, 0, len(c.prefixes))
+	for _, prefix := range c.prefixes {
+		updateStatusFuncs = append(updateStatusFuncs, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:   fmt.Sprintf("%sUpgradeable", prefix),
+			Status: operatorv1.ConditionTrue,
+		}))
+	}
+	return updateStatusFuncs, nil
```
joelanford, line 104: I don't think we should put these hardcoded `Upgradeable=True` status update functions inside a function called `ensureControllersManaged`. It feels like we're mixing concerns there a bit.

I kinda think we should refactor this slightly to accept a slice map of `[]func(context.Context) (operatorv1.OperatorCondition, error)`, where that function (not sure I have the input parameters correct) returns a Condition that we should set on our ClusterOperator.

If we had that setup, then this controller would boil down to:
1. Check the management state
2. If we're managed, iterate the `[]func(context.Context) (Status, error)`, calling each function to build `updateStatusFuncs`
3. Call `v1helpers.UpdateStatus`.

And it would become a generic `ConditionsController`. And if we had _that_ then we could put the `OLMManagedOperatorsUpgradeable` condition logic into a completely separate package/function that only needs to implement that function signature.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,245 @@
...
+	if len(incompatibleOperators) > 0 {
+		updateStatusFuncs = append(updateStatusFuncs, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:    "Upgradeable",
+			Status:  operatorv1.ConditionFalse,
+			Reason:  incompatibleOperatorsInstalled,
+			Message: strings.Join(incompatibleOperators, ","),
+		}))
+	} else {
+		updateStatusFuncs = append(updateStatusFuncs, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:   "Upgradeable",
```
joelanford, line 187: I think there's some helper logic happening somewhere that builds an `Upgradeable` condition for us based on the values of all of the other condition types that have a suffix of `Upgradeable`. We should double check that, but if so, we would need to use something like `OLMManagedOperatorsUpgradeable` for this condition type.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	return props, nil
+}
+
+func (c *dynamicUpgradeableConditionController) buildHelmStore(secretClient v1.SecretInterface) helm.Storage {
+	log := func(s string, args ...interface{}) { c.logger.V(4).Info(fmt.Sprintf(s, args...)) }
+	csConfig := storage.ChunkedSecretsConfig{
+		ChunkSize:      0,
+		MaxReadChunks:  0,
+		MaxWriteChunks: 0,
```
joelanford, line 207: You can just delete these fields from the struct initialization since they are the default values.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,245 @@
...
+		Log:            log,
+	}
+
+	return helm.Storage{
+		Driver:     storage.NewChunkedSecrets(secretClient, "operator-controller", csConfig),
+		MaxHistory: 0,
```
joelanford, line 213: Same here. `MaxHistory` can be removed.

pkg/controller/dynamicupgradeableconditionscontroller.go
joelanford: Either that, or maybe CVO recognizes and aggregates all conditions with the `Upgradeable` suffix as `Upgradeable`.

pkg/controller/dynamicupgradeableconditionscontroller.go
joelanford: For anything _not_ being moved to operator-controller (e.g. `MaxOpenShiftVersionProperty`), we should unexport it since we don't intend for it to be used outside this package.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,245 @@
...
+	"helm.sh/helm/v3/pkg/storage/driver"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+	v1 "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/klog/v2"
+)
+
+const (
+	incompatibleOperatorsInstalled = "IncompatibleOperatorsInstalled"
```
joelanford, line 31: Convention on reason constants is to prefix the identifier with `reason` (or `Reason` if we were exporting it)

```suggestion
	reasonIncompatibleOperatorsInstalled = "IncompatibleOperatorsInstalled"
```

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,245 @@
...
+	current, err := semver.ParseTolerant(raw)
+	if err != nil {
+		return nil, fmt.Errorf("cluster version has invalid desired release version: %w", err)
+	}
+
+	next, err := nextY(current)
+	if err != nil {
+		c.logger.V(4).Error(err, "Error finding next OCP Y-stream release")
+		return nil, err
+	}
```
joelanford, line 124: None of this changes unless the process restarts, right? I'd recommend two things:
1. Do envvar lookups further up the call stack, maybe in `runOperator` in `main.go`?
2. Make a helper function that accepts `raw` and returns `next, err` and also call _that_ function in `main.go`, passing `next` into the code that checks for incompatible operators.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,239 @@
...
+	csConfig := storage.ChunkedSecretsConfig{Log: log}
+
+	return helm.Storage{
+		Driver: storage.NewChunkedSecrets(secretClient, "operator-controller", csConfig),
+		Log:    log,
+	}
+}
+
+const (
+	releaseEnvVar = "RELEASE_VERSION" // OpenShift's env variable for defining the current release
```
joelanford, line 213: re: my earlier comment about moving the lookup out to runOperator, I'm not sure we need this constant at all if we're just doing the lookup in one place. Can probably hardcode os.Getenv("RELEASE_VERSION").

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,239 @@
...
+func (c *dynamicUpgradeableConditionController) ensureControllersManaged() ([]v1helpers.UpdateStatusFunc, error) {
+	c.logger.V(4).Info("verifying state of olm controllers")
+	defer c.logger.V(4).Info("done verifying state of olm controllers")
+	opSpec, _, _, err := c.operatorClient.GetOperatorState()
+	if err != nil {
+		return nil, err
+	}
+	if opSpec.ManagementState != operatorv1.Managed {
+		return nil, nil
+	}
```
joelanford, line 95: Just ran across this code in library-go that seems more robust: https://github.com/openshift/library-go/blob/3c41fd1dea0be2102a6419eb7772bec7ea1b4b38/pkg/operator/staticresourcecontroller/static_resource_controller.go#L287-L289

manifests/0000_51_olm_06_deployment.yaml
```diff
@@ -71,6 +71,8 @@ spec:
           value: quay.io/openshift/origin-olm-catalogd:latest
         - name: OPERATOR_CONTROLLER_IMAGE
           value: quay.io/openshift/origin-olm-operator-controller:latest
+        - name: RELEASE_VERSION
+          value: "0.0.1-snapshot"
```
joelanford, line 75: Oh, I see we already have `OPERATOR_IMAGE_VERSION` set to `0.0.1-snapshot`, which is the magic string that is replaced with the payload version.

I think we should use that.

pkg/controller/dynamicupgradeableconditionscontroller.go
joelanford: I just found that we already have the payload version via `OPERATOR_IMAGE_VERSION`, and library-go has a function that we already use that abstracts it. Look for `status.VersionForOperatorFromEnv()` in `main.go`

We can probably piggyback on all of that. Editing my comment above with this new info.

pkg/controller/dynamicupgradeableconditionscontroller.go
```diff
@@ -0,0 +1,245 @@
...
+)
+
+func nextY(v semver.Version) (semver.Version, error) {
+	v.Build = nil                // Builds are irrelevant
+	v.Pre = nil                  // Next Y release
+	return v, v.IncrementMinor() // Sets Y=Y+1 and Z=0
+}
+
+func toSemver(max string) (*semver.Version, error) {
+	value := strings.Trim(max, "\"")
```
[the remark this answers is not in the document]
gallettilance: `ParseTolerant` does this

manifests/0000_51_olm_06_deployment.yaml
```diff
@@ -58,7 +58,7 @@ spec:
         - /cluster-olm-operator
         args:
         - start
-        - -v=2
+        - -v=4
```
joelanford, line 61: I _think_ this can be fully removed now that the logging controller is added. The verbosity can now be managed via the `OLM.spec.operatorLogLevel` field.

cmd/cluster-olm-operator/main.go
```diff
@@ -105,13 +106,28 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
...
+
 	upgradeableConditionController := controller.NewStaticUpgradeableConditionController(
 		"OLMStaticUpgradeableConditionController",
 		cl.OperatorClient,
 		cc.EventRecorder.ForComponent("OLMStaticUpgradeableConditionController"),
 		controllerNames,
 	)
 
+	incompatibleOperatorController := controller.NewIncompatibleOperatorController(
+		"OLMincompatibleOperatorController",
```
joelanford, line 123:
```suggestion
		"OLMIncompatibleOperatorController",
```

I know `OLMI` is a bit jarring since we're combining an acronym with CamelCase, but that is the convention (e.g. `IPAddress`, `TCPAddr`, etc.)

pkg/clients/clients.go
```diff
@@ -91,23 +94,40 @@ func New(cc *controllercmd.ControllerContext) (*Clients, error) {
 		return nil, err
 	}
 
+	infFact := dynamicinformer.NewDynamicSharedInformerFactory(dynClient, defaultResyncPeriod)
+
+	clusterExtensionGVR := schema.GroupVersionResource{
+		Group:    ocv1alpha1.ClusterExtensionGVK.Group,
+		Version:  ocv1alpha1.ClusterExtensionGVK.Version,
+		Resource: "clusterextensions",
+	}
```
joelanford, line 103: Pretty sure this GVK variable is already gone from upstream. I'd suggest going ahead and changing this to the following:
```suggestion
 	clusterExtensionGVR := ocv1alpha1.GroupVersion.WithResource("clusterextensions")
```

pkg/controller/incompatibleOperatorController.go
joelanford, line 1: Nit: Go filenames are usually snake case. 

```
pkg/controller/incompatible_operator_controller.go
```

pkg/controller/incompatibleOperatorController.go
```diff
@@ -0,0 +1,154 @@
...
+		operatorClient:         operatorClient,
+		logger:                 klog.FromContext(context.TODO()).WithName(name),
+	}
+
+	return factory.New().WithSync(c.sync).WithSyncDegradedOnError(operatorClient).WithInformers(operatorClient.Informer(), clusterExtensionClient.Informer().Informer()).ToController(name, eventRecorder)
+}
+
+func (c *incompatibleOperatorController) sync(ctx context.Context, _ factory.SyncContext) error {
+	c.logger.V(4).Info("sync started")
+	defer c.logger.V(4).Info("done syncing")
```
joelanford, line 54: I would put these at standard `c.logger.Info` level.

Also nit: `s/done syncing/sync ended/` just for symmetry with "sync started"

pkg/controller/incompatibleOperatorController.go
```diff
@@ -0,0 +1,154 @@
...
+
+	updateStatusFuncs, err := c.checkIncompatibleOperators()
+	if err != nil {
+		c.logger.V(4).Error(err, "error checking incompatible operators")
+		return err
```
joelanford, line 59: If we return an error here without logging it ourselves, does something else log it?

If so, I think we can drop the `c.logger.V(4).Error(...)` logs.
If not, we should keep these, but change to `c.logger.Error(...)` to make clear that we always intend for these to be logged.

pkg/controller/incompatibleOperatorController.go
```diff
@@ -0,0 +1,154 @@
...
+			Reason:  utils.ReasonIncompatibleOperatorsInstalled,
+			Message: strings.Join(incompatibleOperators, ","),
+		}))
+	} else {
+		updateStatusFuncs = append(updateStatusFuncs, v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:   "InstalledOLMOperatorsUpgradeable",
+			Status: operatorv1.ConditionTrue,
+		}))
+	}
+	return updateStatusFuncs, nil
```
joelanford, line 135: Nit: move this back out of this function, and rename the function to `getIncompatibleOperators`? Then have it return the sorted list of incompatible operators?

joelanford: Then in sync, we could construct the `operatorv1.OperatorCondition` we want, and then finally call `v1helpers.UpdateStatus` with an inline `v1helpers.UpdateConditionFn`.

No need for carry around `updateStatusFuncs`.

pkg/utils/utils.go
joelanford, line 1: I'd put `utils` in the `internal` directory

pkg/utils/utils.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+const (
+	ReasonIncompatibleOperatorsInstalled = "IncompatibleOperatorsInstalled"
+	MaxOpenShiftVersionProperty          = "olm.maxOpenShiftVersion"
+	OwnerKindKey                         = "olm.operatorframework.io/owner-kind"
+	OwnerNameKey                         = "olm.operatorframework.io/owner-name"
+	PackageNameKey                       = "olm.operatorframework.io/package-name"
+	BundleNameKey                        = "olm.operatorframework.io/bundle-name"
+	BundleVersionKey                     = "olm.operatorframework.io/bundle-version"
```
joelanford, line 17: I'd move these back to the incompatibleOperatorController file and unexport them all.

pkg/controller/incompatibleOperatorController.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+func NewIncompatibleOperatorController(name string, nextOCPMinorVersion *semver.Version, kubeclient kubernetes.Interface, clusterExtensionClient *clients.ClusterExtensionClient, operatorClient *clients.OperatorClient, eventRecorder events.Recorder) factory.Controller {
+	c := &incompatibleOperatorController{
+		name:                   name,
+		nextOCPMinorVersion:    nextOCPMinorVersion,
+		kubeclient:             kubeclient,
+		clusterExtensionClient: clusterExtensionClient,
+		operatorClient:         operatorClient,
+		logger:                 klog.FromContext(context.TODO()).WithName(name),
```
joelanford, line 46: Do we have a logger we could pass in rather than constructing from `klog.FromContext`? It feels a little weird to call that from a function what will never expect a context.

If not, maybe there's a different `klog` function that could just hand us a new logger directly rather than referencing through a context that we know won't already have one?

pkg/utils/utils.go
gallettilance: can't because it's used in main.go

joelanford: `main.go` should be able to use stuff inside `internal`.

pkg/controller/dynamicupgradeableconditionscontroller.go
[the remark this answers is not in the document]
joelanford: `ParseTolerant` accepts _more_ than just that though. It accepts 3 and 4 as well. Assuming 2 dominates usage in the catalogs, I think we should reject 1, 3, and 4, and then proactively tell authors using 1, 3, and 4 that they need to switch to 2.

Like I said before, it would be good to have some actual data here. But I don't think it helps anyone for there to be 4 ways to represent the same thing. That just leads to confusion and bugs.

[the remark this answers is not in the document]
joelanford: I just searched the existing catalogs and found only 1 and 2. Let's reject 3 and 4 for now and then come back to the question of whether to also drop 1 or 2?

internal/utils/utils.go
```diff
@@ -0,0 +1,33 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	semver "github.com/blang/semver/v4"
+)
+
+func GetNextOCPMinorVersion(versionString string) (*semver.Version, error) {
+	v, err := ToSemver(versionString)
```
joelanford, line 11: Parsing the OCP version should use strict semver parsing, not our tolerant "maxOpenshiftVersion" parsing.

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,168 @@
...
+	return factory.New().WithSync(c.sync).WithSyncDegradedOnError(operatorClient).WithInformers(operatorClient.Informer(), clusterExtensionClient.Informer().Informer()).ToController(name, eventRecorder)
+}
+
+func (c *incompatibleOperatorController) sync(ctx context.Context, _ factory.SyncContext) error {
+	c.logger.Info("sync started")
+	defer c.logger.Info("sync finished")
+
+	incompatibleOperators, err := c.getIncompatibleOperators()
+	if err != nil {
+		c.logger.V(4).Error(err, "error checking incompatible operators")
```
joelanford, line 68: Seems like we should propagate this error to the Operator status? Probably by setting `InstalledOLMOperatorsUpgradeable=Unknown` in this case?

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,168 @@
...
+
+	store := c.buildHelmStore(c.kubeclient.CoreV1().Secrets("openshift-operator-controller"))
+
+	// Get all ClusterExtensions incompatible with next Y-stream
+	for _, obj := range ceList {
+		ce, ok := obj.(*unstructured.Unstructured)
+		if !ok {
+			c.logger.V(4).Error(nil, fmt.Sprintf("Unable to Cast Object to unstructured.Unstructured. Object is of type %T", obj))
+			return nil, nil
+		}
```
joelanford, line 111: Do we use anything other than `ce.GetName()`? If not, no need for this type cast. We can use `obj.GetName()` directly, I think.

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,168 @@
...
+		if rel.Chart == nil || rel.Chart.Metadata == nil {
+			continue
+		}
+		props, err := propertyListFromPropertiesAnnotation(rel.Chart.Metadata.Annotations["olm.properties"])
+		if err != nil {
+			c.logger.V(4).Error(err, "Error converting olm.properties")
+			return nil, err
+		}
+		for _, p := range props {
+			if p.Type == maxOpenShiftVersionProperty {
```
joelanford, line 132: We probably need to count the number of times we see this property and emit an error if we see it more than once.

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,168 @@
...
+				version := string(p.Value)
+				maxOCPVersion, err := utils.ToAllowedSemver(version)
+				if err != nil {
+					c.logger.V(4).Info(fmt.Sprintf("Error converting to semver for version %s: %v", version, err))
+					continue
```
joelanford, line 137: Seems like this error should be propagated back up to the caller.

We should probably collect a slice of errors for each clusterExtension, and then aggregate all of _those_ errors up into an error that we return from this function. So we can `errors.Join` to get a single error per cluster extension, and then `errors.Join` again to get a single error that covers all cluster extensions.

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,168 @@
...
+				}
+				if maxOCPVersion != nil && !maxOCPVersion.GTE(*c.nextOCPMinorVersion) {
+					// Incompatible
+					incompatibleOperators = append(incompatibleOperators, rel.Labels[bundleNameKey])
```
joelanford, line 141: IMO, we should include the cluster extension name in the message to help users identify the object that needs to change in order to resolve this issue.

Something like:
```
fmt.Sprintf("bundle %q for ClusterExtension %q", rel.Labels[bundleNameKey], ce.GetName())

File: pkg/controller/incompatible_operator_controller.go [hunk 43]
```diff
@@ -0,0 +1,168 @@
...
+		}
+		for _, p := range props {
+			if p.Type == maxOpenShiftVersionProperty {
+				version := string(p.Value)
+				maxOCPVersion, err := utils.ToAllowedSemver(version)
+				if err != nil {
+					c.logger.V(4).Info(fmt.Sprintf("Error converting to semver for version %s: %v", version, err))
+					continue
+				}
+				if maxOCPVersion != nil && !maxOCPVersion.GTE(*c.nextOCPMinorVersion) {
```
joelanford, line 139: Should this be strictly less than?
- `maxOCPVersion=4.15`, `nextMinor=4.15` => all good
- `maxOCPVersion=4.16`, `nextMinor=4.15` => all good
- `maxOCPVersion=4.15`, `nextMinor=4.16` => incompatible

pkg/controller/incompatible_operator_controller.go
joelanford: (for a single bundle)

pkg/controller/incompatible_operator_controller.go
gallettilance: yeah not greater or equal is strictly less

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		name := metaObj.GetName()
+		rel, err := store.Deployed(name)
+		if errors.Is(err, driver.ErrNoDeployedReleases) {
+			c.logger.Info("Cluster Extension not yet deployed - will check again later")
+			continue
+		}
+		if err != nil {
+			c.logger.V(4).Error(err, "Error returning the last deployed release")
+			return nil, err
```
joelanford, line 127: Append to `errs` and continue if we fail to get deployed release for this CE?

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+
+		if rel.Chart == nil || rel.Chart.Metadata == nil {
+			continue
+		}
+		props, err := propertyListFromPropertiesAnnotation(rel.Chart.Metadata.Annotations["olm.properties"])
+		if err != nil {
+			c.logger.V(4).Error(err, "Error converting olm.properties")
+			return nil, err
+		}
```
joelanford, line 137: Same here.

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,181 @@
...
+		numMaxOCPProps := 0
+		for _, p := range props {
+			if p.Type == maxOpenShiftVersionProperty {
+				numMaxOCPProps++
+				maxOCPVersion, err := utils.ToAllowedSemver(string(p.Value))
+				if err != nil {
+					errs = append(errs, fmt.Errorf(fmt.Sprintf("Error converting to semver for version %s: %v", string(p.Value), err)))
```
joelanford, line 144: We should probably tell the user what the ClusterExtension and bundle name were for this error.

pkg/controller/incompatible_operator_controller.go
joelanford: Doh! I missed the `!`

joelanford: Probably still a nit for me to use `LT` without the negation, but non-blocking.

pkg/controller/incompatible_operator_controller.go
joelanford: Also, we can't really say that this particular error is because of `InvalidBundleProperties`, so we should probably come up with an even more generic reason:

- `UnexpectedFailure`
- `FailureGettingExtensionMetadata`

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,181 @@
...
+
+func propertyListFromPropertiesAnnotation(raw string) ([]property.Property, error) {
+	var props []property.Property
+	if err := json.Unmarshal([]byte(raw), &props); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal properties annotation: %w", err)
+	}
+	return props, nil
+}
+
+func (c *incompatibleOperatorController) buildHelmStore(secretClient v1.SecretInterface) helm.Storage {
```
joelanford, line 173: Non-blocking nit: Any where we use `Operator`, use `Extension` instead.

cmd/cluster-olm-operator/main.go
```diff
@@ -105,13 +106,28 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		deploymentControllerList = append(deploymentControllerList, controller)
 	}
 
+	operatorImageVersion := status.VersionForOperatorFromEnv()
```
joelanford, line 109: Re: my other comment about terminology: this would still be "operator" since `cluster-olm-operator` _is_ an operator

internal/utils/utils.go
```diff
@@ -0,0 +1,48 @@
...
+	if err != nil {
+		return &v, err
+	}
+	v.Build = nil                 // Builds are irrelevant
+	v.Pre = nil                   // Next Y release
+	return &v, v.IncrementMinor() // Sets Y=Y+1 and Z=0
+}
+
+func ToAllowedSemver(value string) (*semver.Version, error) {
+	if value == "" {
```
joelanford, line 22: This doesn't look like it'll work for this scenario:
```go
value := `"4.16"`
vers, err := ToAllowedSemver(value)
```

To fix, I think I'd suggest parsing as JSON, something like the following. I don't think it's a good idea to use naive prefix and suffix trimming of double quotes because there are likely JSON complexities that that method would gloss over.

```go
var rawV interface{}
if err := json.Unmarshal(&rawV, value); err != nil {
      return nil, err
}
switch val := rawV.(type) {
case string:
    // handle string
case <whateverTheRightFloatTypeIs>:
    // handle float, but need to think about weird rounding that might happen
    // maybe we see that it's a float, but then parse the raw value like you already are?
default:
      return nil, fmt.Errorf("invalid type %T for olm.maxOpenshiftVersion: %s", string(value))
}
```

Probably makes sense to add some unit tests here to make sure we have our bases covered.

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	defer c.logger.Info("sync finished")
+
+	var updateStatusFn v1helpers.UpdateStatusFunc
+	incompatibleOperators, err := c.getIncompatibleOperators()
+	if len(incompatibleOperators) > 0 {
+		updateStatusFn = v1helpers.UpdateConditionFn(operatorv1.OperatorCondition{
+			Type:    typeIncompatibelOperatorsUpgradeable,
+			Status:  operatorv1.ConditionFalse,
+			Reason:  reasonIncompatibleOperatorsInstalled,
+			Message: strings.Join(incompatibleOperators, ","),
```
joelanford, line 75: Provide more context in this message? I know in the past users have been confused and think that patch versions are blocked because of this. 
```
fmt.Sprintf("Found ClusterExtensions that require upgrades prior to upgrading cluster to version %d.%d: %s", nextOCPVersion.Major, nextOCPVersion.Minor, strings.Join(incompatibleOperators, ","))
```

joelanford: Also, I think that if `err != nil` here, we should still include it's message to give users that context as early as possible. Like "Your cluster has incompatible operators installed, so Upgradeable=False.... but also you've got these other problems that will need your attention too"

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,181 @@
...
+		for _, p := range props {
+			if p.Type == maxOpenShiftVersionProperty {
+				numMaxOCPProps++
+				maxOCPVersion, err := utils.ToAllowedSemver(string(p.Value))
+				if err != nil {
+					errs = append(errs, fmt.Errorf(fmt.Sprintf("Error converting to semver for version %s: %v", string(p.Value), err)))
+					continue
+				}
+				if numMaxOCPProps > 1 {
+					errs = append(errs, fmt.Errorf("%s has more than one %s", name, maxOpenShiftVersionProperty))
```
joelanford, line 148: Same here about including ClusterExtension and bundle name context in the error.

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,181 @@
...
+	var errs []error
+	// Get all ClusterExtensions incompatible with next Y-stream
+	for _, obj := range ceList {
+		metaObj, ok := obj.(metav1.Object)
+		if !ok {
+			errs = append(errs, fmt.Errorf("metav1.Object type assertion failed for object %v", obj))
+			continue
+		}
+		name := metaObj.GetName()
+		rel, err := store.Deployed(name)
```
joelanford, line 120:
```
logger := c.logger.WithValues("clusterExtensionName", metaObj.GetName())
```

And then use `logger` everywhere inside the loop?

pkg/controller/incompatible_operator_controller.go
```diff
@@ -0,0 +1,181 @@
...
+			continue
+		}
+		if err != nil {
+			c.logger.V(4).Error(err, "Error returning the last deployed release")
+			return nil, err
+		}
+
+		if rel.Chart == nil || rel.Chart.Metadata == nil {
+			continue
+		}
```
joelanford, line 132: Log this? It would be pretty unexpected, I think.

PR #70, opened 2024-09-26
honor "operatorLogLevel" from the operator spec

joelanford:
Right now the `spec.operatorLogLevel` and `spec.logLevel` values in the operator spec are ignored.

This PR adds a logging controller that ensures that at least the operator log level is dynamically updated based on the value of `spec.operatorLogLevel`

NOTE: handling `spec.logLevel` will take a bit more time, as it will likely require some coordination with catalogd and operator-controller maintainers.


joelanford:
@gallettilance, hopefully this makes your debugging at least a little bit easier. It'll definitely be nice for support associates and customers.

cmd/cluster-olm-operator/main.go
```diff
@@ -124,9 +125,11 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		cc.EventRecorder.ForComponent("olm"),
 	)
 
+	operatorLoggingController := loglevel.NewClusterOperatorLoggingController(cl.OperatorClient, cc.EventRecorder.ForComponent("ClusterOperatorLoggingController"))
```
tmshort, line 128: Why does this feel like it needs to have "OLM" in it? (Like the other `ForComponents`?)

pkg/controller/builder.go
```diff
@@ -91,6 +93,7 @@ func (b *Builder) BuildControllers(subDirectories ...string) (map[string]factory
 					b.Clients.KubeInformerFactory.Apps().V1().Deployments(),
 					nil,
 					[]deploymentcontroller.ManifestHookFunc{
+						replaceVerbosityHook("${LOG_VERBOSITY}"),
```
tmshort, line 96: Is this "variable" used anywhere? Is it somewhere in our manifests?

cmd/cluster-olm-operator/main.go
joelanford: Ah. I'll make it match. Good catch. My thinking at the time was "this is already scoped to the `openshift-cluster-olm-operator` namespace, but since the other components do it, we should be consistent.

pkg/controller/builder.go
joelanford: It will be. This needs to be here and doing replacements before it shows up in our manifests.

vendor/github.com/openshift/library-go/pkg/operator/loglevel/util.go
```diff
@@ -0,0 +1,103 @@
...
+	case operatorv1.Normal:
+		return 2
+	case operatorv1.Debug:
+		return 4
+	case operatorv1.Trace:
+		return 6
+	case operatorv1.TraceAll:
+		return 8
+	default:
+		return 2
```
perdasilva, line 25: I agree that we probably shouldn't throw an error error. But, I'm thinking it might be worthwhile if we log it, just to help us understand a possible future state where a new log level is added, used, but doesn't manifest itself (as unlikely that is likely to be XD)

vendor/github.com/openshift/library-go/pkg/operator/loglevel/util.go
```diff
@@ -0,0 +1,103 @@
...
+		if flagValue, ok := f.Value.(*klog.Level); ok {
+			level = flagValue
+		}
+	}
+
+	// Second, if the '-v' was not set but is still present in flags defined for the command, attempt to acquire it
+	// by visiting all flags.
+	if level == nil {
+		flag.VisitAll(func(f *flag.Flag) {
+			if level != nil {
```
perdasilva, line 82: Could we remove this if and add the return in the next if?

vendor/github.com/openshift/library-go/pkg/operator/loglevel/util.go
perdasilva: Or maybe even throw an error and publish an event? Then the decision to use Normal for unknown is a bit higher up the stack?

vendor/github.com/openshift/library-go/pkg/operator/loglevel/util.go
joelanford: This is not our code, coming in from vendored library-go.

vendor/github.com/openshift/library-go/pkg/operator/loglevel/util.go
joelanford: Sounds reasonable! Should we make a PR in library-go for this? We could also write our own helper, but then we'd be maintaining it.

perdasilva: I can poke at that. Been cooking up one with k8s lib update for OpenShift/api to help clean up Lance's PR

vendor/github.com/openshift/library-go/pkg/operator/loglevel/util.go
perdasilva: Sorry - totally missed that it was vendored =S

PR #72, opened 2024-10-09
OCPBUGS-42860: degraded co operator

gallettilance:


tmshort:
Ugh... the e2e job is broken (known issue), so we gotta wait for that to be fixed, or get an override.

PR #73, opened 2024-10-11
OPRUN-3571: (feature): add clustercatalog controller for managing default clustercatalogs

everettraven:


pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,410 @@
...
+	}
+
+	cond := meta.FindStatusCondition(catalog.Status.Conditions, catalogdv1alpha1.TypeServing)
+	if cond == nil {
+		return fmt.Errorf("could not find status condition type %q", catalogdv1alpha1.TypeServing)
+	}
+
+	// TODO: should we considered observed generation?
+	// if Serving condition observedGeneration = 1 but meta.generation = 6 is that "healthy"
+	// because $something is being served?
```
everettraven, line 397: Specifically calling this TODO out for reviewers - please share your thoughts!

pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,401 @@
...
+		return nil
+	}
+
+	// Generate and apply a patch for the diff between original and modified
+	diffPatch, err := generatePatch(original, modified)
+	if err != nil {
+		return fmt.Errorf("generating patch: %w", err)
+	}
+
+	_, err = a.client.Resource(a.gv.WithResource(a.resource)).Patch(ctx, original.GetName(), types.MergePatchType, diffPatch, metav1.PatchOptions{
```
everettraven, line 156: Note: After a discussion with @joelanford I'm going to spend some more time looking to see if we can simply use Server Side Apply instead of attempting to do a merge patch.

I originally ran into a swath of issues trying to do SSA but Joe shared a couple more resources with me that may be helpful in resolving some of the issues I encountered.

pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,401 @@
...
+	"k8s.io/utils/ptr"
+
+	catalogdv1alpha1 "github.com/operator-framework/catalogd/api/core/v1alpha1"
+)
+
+const (
+	reasonCreated    = "Created"
+	reasonPatched    = "Patched"
+	reasonAsExpected = "AsExpected"
+)
```
LalatenduMohanty, line 35: We should add some Go doc comments around why we have these reasons (constants).

pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,401 @@
...
+
+func NewClusterCatalogController(name string, manifest []byte, operatorClient *clients.OperatorClient, dynamicClient dynamic.Interface, clusterCatalogClient *clients.ClusterCatalogClient, recorder events.Recorder) (factory.Controller, error) {
+	obj, _, err := scheme.Codecs.UniversalDecoder().Decode(manifest, nil, &unstructured.Unstructured{})
+	if err != nil {
+		return nil, fmt.Errorf("decoding manifest data: %w", err)
+	}
+
+	uObj := obj.(*unstructured.Unstructured)
```
LalatenduMohanty, line 43: These code pattern looks different than the other controllers we have e.g. pkg/controller/incompatible_operator_controller.go or pkg/controller/staticupgradeableconditioncontroller.go . Just to make sure this is an explicit choice.

pkg/controller/clustercatalog_controller.go
LalatenduMohanty: +1 to server side apply.

manifests/0000_51_olm_02_operator_clusterrole.yaml
```diff
@@ -219,6 +219,17 @@ rules:
     - get
     - list
     - watch
+  - apiGroups:
+    - olm.operatorframework.io
+    resources:
+    - clustercatalogs
+    verbs:
+    - get
```
everettraven, line 227: We probably don't need get permissions since we only ever fetch from the cache.

manifests/0000_51_olm_02_operator_clusterrole.yaml
```diff
@@ -219,6 +219,17 @@ rules:
...
+    - list
+    - watch
+    - create
+    - update
```
everettraven, line 231: If a server-side apply patch request results in creating or updating the resource do we need these permissions? I don't think we do, but I'll try to verify tomorrow when I can get another clusterbot cluster spun up

go.mod
```diff
@@ -139,3 +140,5 @@ require (
 	sigs.k8s.io/structured-merge-diff/v4 v4.4.1 // indirect
 	sigs.k8s.io/yaml v1.4.0 // indirect
 )
+
+replace github.com/operator-framework/catalogd => github.com/openshift/operator-framework-catalogd v0.0.0-20241018021036-4275eb2ee38d
```
joelanford, line 144: Is this temporary? If not how do we place to keep this updated?

manifests/0000_51_olm_02_operator_clusterrole.yaml
joelanford: Pretty sure `update` can be removed. I can't remember if `create` is required even if a `patch` operation performs the create.

pkg/controller/builder.go
```diff
@@ -102,10 +104,27 @@ func (b *Builder) BuildControllers(subDirectories ...string) (map[string]factory
 				return nil
 			}
 
+			if manifestGVK.Kind == "ClusterCatalog" && manifestGVK.Group == "olm.operatorframework.io" {
+				controllerName := controllerNameForObject(namePrefix, &manifest)
```
joelanford, line 108: Any reason we need an individual controller for each `ClusterCatalog`? Seems like we should be able to have a single controller that handles all of them.

pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			return false, err
+		}
+		return management.IsOperatorManaged(operatorSpec.ManagementState) && (operatorSpec.ManagementState != operatorv1.Removed || management.IsOperatorNotRemovable()), nil
+	}
+}
+
+func defaultShouldUpdateFunc() shouldUpdateFunc {
+	return func(manifest []byte, u *unstructured.Unstructured) (bool, error) {
+		required, _, err := scheme.Codecs.UniversalDecoder().Decode(manifest, nil, &unstructured.Unstructured{})
```
joelanford, line 71: I'm having trouble groking the unstructured objects.
1. Can you give `u` a more descriptive name?
2. Maybe make a variable (or add a comment) for the `&unstructured.Unstructured{}` that we pass to `Decode` so that it is more obvious what that parameter is for.

pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,156 @@
...
+	logger := klog.FromContext(ctx).WithName(c.name)
+	logger.V(4).Info("sync started")
+	defer logger.V(4).Info("sync finished")
+
+	managed, err := c.managedFunc()
+	if err != nil {
+		return fmt.Errorf("checking if operator is managed: %w", err)
+	}
+	if !managed {
+		return nil
```
joelanford, line 131: Log that the operator is unmanaged?

pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,156 @@
...
+// Kubernetes resource YAML, and compares it to an *unstructured.Unstructured.
+// if there is a difference between the raw bytes and the *unstructured.Unstructured
+// that warrants an update to the *unstructured.Unstructured, this function will return
+// true with a nil error. Any errors encountered during the evaluation will be returned.
+type shouldUpdateFunc func([]byte, *unstructured.Unstructured) (bool, error)
+
+// getCatalogFunc is a function that gets a ClusterCatalog resource using the provided
+// key. It is returned as a runtime.Object. Any errors encountered during the process
+// of fetching the ClusterCatalog resource will be returned.
+type getCatalogFunc func(types.NamespacedName) (runtime.Object, error)
```
joelanford, line 102: Since this is already an abstraction, maybe make this return a concrete type?
```suggestion
type getCatalogFunc func(types.NamespacedName) (*unstructured.Unstructured, error)
```

pkg/controller/clustercatalog_controller_test.go
```diff
@@ -0,0 +1,127 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func TestClusterCatalogControllerSync(t *testing.T) {
+	for _, tc := range []struct {
+		name        string
+		ctrl        *clusterCatalogController
+		shouldError bool
```
joelanford, line 18: I'd suggest changing this to the following:

```suggestion
		assertError func(t *testing.T, err error)
```

That way we don't actually assert on a specific error, it would be possible to accidentally introduce a different code path that errors in a different, unexpected way and for these tests to still pass.

Example of where I learned this lesson: https://github.com/operator-framework/operator-registry/pull/709

pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,156 @@
...
+	shouldUpdate, err := c.shouldUpdateFunc(c.manifest, catalog.(*unstructured.Unstructured))
+	if err != nil {
+		return fmt.Errorf("determining if clustercatalog %q should be updated: %w", c.key, err)
+	}
+
+	if !shouldUpdate {
+		return nil
+	}
+
+	return c.applyFunc(
```
joelanford, line 148: Standard info level log for when we apply a patch?

pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,156 @@
...
+		return fmt.Errorf("fetching clustercatalog %q: %w", c.key, err)
+	}
+
+	shouldUpdate, err := c.shouldUpdateFunc(c.manifest, catalog.(*unstructured.Unstructured))
+	if err != nil {
+		return fmt.Errorf("determining if clustercatalog %q should be updated: %w", c.key, err)
+	}
+
+	if !shouldUpdate {
+		return nil
```
joelanford, line 145: Debug log for when we see that no update is necessary?

pkg/controller/clustercatalog_controller.go
```diff
@@ -0,0 +1,156 @@
...
+	applyFunc        applyFunc
+	managedFunc      managedFunc
+	shouldUpdateFunc shouldUpdateFunc
+	catalogGetFunc   getCatalogFunc
+}
+
+func (c *clusterCatalogController) sync(ctx context.Context, _ factory.SyncContext) error {
+	logger := klog.FromContext(ctx).WithName(c.name)
+	logger.V(4).Info("sync started")
+	defer logger.V(4).Info("sync finished")
```
joelanford, line 124: Make these logs be emitted with default logging configuration?

go.mod
everettraven: My thinking was that this would need to be kept around, but I realize we also didn't do this replacement for the operator-controller dependency we have.

Is there a particular reason we feel comfortable using the upstream vs a specific commit in the corresponding downstream repo? I would expect us to want to ensure everything downstream is properly synced and referencing the corresponding commits for the same OCP version. If we use the upstream and make a downstream only change to something like our group (i.e going from `olm.operatorframework.io` to `olm.openshift.io`) cluster-olm-operator would break.

I do realize this is likely to get stale pretty quickly, but for keeping it up to date I would expect us to update our downstream sync tooling to also know how and when to update these replacements such that everything is in sync for our downstream uses. For backporting fixes, this would likely require manual updates.

pkg/controller/builder.go
everettraven: We certainly could have a single controller that handles all of them. I'm not _super_ opinionated here, but my thinking for individual controller was that we have:
- Slightly easier to understand logic in the controller because we don't have to loop through all of them and have special logic for handling failure to keep resources updated in the middle of the loop
- More nuanced signal for which `ClusterCatalog` resources are degraded because each controller gets it's own `Degraded` condition. If we went with a single controller any failure would result in a `Degraded` signal that requires parsing the string message to understand which `ClusterCatalog` resources are actually degraded. Those that care about the health of individual `ClusterCatalog` resources could build better automation against the individual degraded conditions.
- Maintains consistency with the `DeploymentController` that we use for specific resources (although this is the least important point IMO)

I felt that these pros outweighed the con of having more controllers running, but provided a strong argument for having a single controller that proves otherwise I'm happy to change this.

pkg/controller/clustercatalog_controller.go
everettraven: I can give `u` a more descriptive name. 

For point 2, see https://pkg.go.dev/k8s.io/apimachinery@v0.31.1/pkg/runtime#Decoder for more information on the inputs and outputs. The TLDR is that the `&unstructured.Unstructured{}` is the target type for the returned `runtime.Object` and may or may not be populated. I'll add a comment with more information.

pkg/controller/clustercatalog_controller.go
everettraven: In this particular case, I generally prefer to use the lowest primitive type and put any type conversion responsibilities on the caller. This makes this code the most extensible because AFAIK all Go types that are used to represent a Kubernetes resource implement the `runtime.Object` interface meaning that you could swap in and out any client that returns a Go type for a Kubernetes resource.

I do think that I should go back and make the signature of the `shouldUpdateFunc` type to take in a `runtime.Object` instead of an `*unstructured.Unstructured` for a similar reason.

pkg/controller/clustercatalog_controller.go
everettraven: What is the "default logging configuration"? I took this logic from existing controllers we have. Is there any other references/documentation to how things should be logged to follow our logging configurations?

pkg/controller/clustercatalog_controller.go
everettraven: Sure. What level would you like this to be logged at?

go.mod
everettraven: Consensus is cross this bridge when we get to it. Will remove this replacement.

grokspawn: I think the 'why upstream versus downstream' is a larger conversation, but without the tools already updated to implement whatever we agree we need to do, this replace seems premature.

pkg/controller/clustercatalog_controller.go
everettraven: Should be done in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676

pkg/controller/clustercatalog_controller.go
everettraven: Should be done in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676

pkg/controller/clustercatalog_controller.go
everettraven: Should be done in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676

pkg/controller/clustercatalog_controller_test.go
everettraven: Should be done in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676

pkg/controller/clustercatalog_controller.go
everettraven: This should be addressed by the more generic approach taken in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676

pkg/controller/clustercatalog_controller.go
everettraven: Should be done in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676

pkg/controller/clustercatalog_controller.go
everettraven: Updated in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676

pkg/controller/builder.go
everettraven: Left as is based on a synchronous discussion.

go.mod
everettraven: Removed in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676

manifests/0000_51_olm_02_operator_clusterrole.yaml
everettraven: update removed in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676 - create had to stay.

manifests/0000_51_olm_02_operator_clusterrole.yaml
everettraven: removed in https://github.com/openshift/cluster-olm-operator/pull/73/commits/5b237d4e4d462f19ff7571661b839393bde6b676

pkg/controller/clustercatalog_controller.go
everettraven: These no longer exist

pkg/controller/clustercatalog_controller.go
everettraven: This code is now completely different

pkg/controller/clustercatalog_controller.go
everettraven: This logic was removed

pkg/controller/dynamicrequiredmanifest_controller_test.go
```diff
@@ -0,0 +1,163 @@
...
+
+func noError() func(t *testing.T, err error) {
+	return func(t *testing.T, err error) {
+		if err != nil {
+			t.Fatalf("expected no error but received error %v", err)
+		}
+	}
+}
+
+func TestClusterCatalogControllerSync(t *testing.T) {
```
everettraven, line 34: Need to update this test name and any references to catalogs

everettraven: Done in https://github.com/openshift/cluster-olm-operator/pull/73/commits/1dd7f1fd8255a006a832402bb6d44726983373d3

cmd/cluster-olm-operator/main.go
```diff
@@ -161,6 +167,13 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		}(c)
 	}
 
+	for _, c := range clusterCatalogControllerList {
+		go func(c factory.Controller) {
+			defer runtime.HandleCrash()
+			c.Run(ctx, 1)
+		}(c)
+	}
```
joelanford, line 175: We put this after the sleep so that the `ClusterCatalog` CRD is established before we attempt to interact with the API?

pkg/clients/clients.go
```diff
@@ -122,6 +117,7 @@ func (c *Clients) StartInformers(ctx context.Context) {
 	c.ConfigInformerFactory.Start(ctx.Done())
 	c.OperatorInformers.Start(ctx.Done())
 	c.ClusterExtensionClient.factory.Start(ctx.Done())
+	c.ClusterCatalogClient.factory.Start(ctx.Done())
```
joelanford, line 120: Have you seen any issues with this? We call `StartInformers` in `main.go` before the `ClusterCatalog` CRD exists on the cluster, and therefore before the API exists.

Is there some sort of retry logic that enables the ClusterCatalog informer to successfully start up after the API shows up?

pkg/controller/dynamicrequiremanifest_controller.go
```diff
@@ -0,0 +1,181 @@
...
+		required, _, err := scheme.Codecs.UniversalDecoder().Decode(manifest, nil, &unstructured.Unstructured{})
+		if err != nil {
+			return false, fmt.Errorf("decoding manifest: %w", err)
+		}
+
+		// we want the inverse of deep derivative to dictate if we should update.
+		// i.e If it is not deep derivative, meaning the requirements are not met, we should update
+		return !equality.Semantic.DeepDerivative(required.(*unstructured.Unstructured).UnstructuredContent(), existingUnstructured), nil
+	}
+}
```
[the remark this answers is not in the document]
everettraven: I can certainly add some more, but I feel the behavior is pretty straightforward for most of them and would prefer not mocking out Kubernetes clients as much as possible, namely the `dynamic.Interface` and `clients.OperatorClient`. That being said we could require the input to be minimal interfaces explicitly for our use cases. I think the most reasonable to add unit testing for would be the `unstructuredShouldUpdateFunc` function because it doesn't actually perform any calls to Kube and would be much more straightforward to test.

I'm happy to make any updates you feel necessary in terms of testing. I know you mention this is non-blocking, but I think we might as well make any additions we feel are important, including unit tests, to this code now rather than later (because we may never do it later).

pkg/clients/clients.go
everettraven: I have not seen any issues with this. I'm not sure if there is logic that retries under the hood. That being said, I've _personally_ only ever tested this in an environment where those CRDs had already been created. I do see that CI is green which seems to track that this is OK, although we don't have any `ClusterCatalog` resources we are managing yet.

everettraven: I'll do some additional digging to see if there is a need to take a different approach

everettraven: It does look like the `Reflector` for an informer will continue to List/Watch with exponential backoff on errors based on https://github.com/kubernetes/client-go/blob/c5146a9031d2c0dbe2a67dede62361ecf3d651bf/tools/cache/reflector.go#L309-L320

I think it's probably safe to leave as-is here with the intention that we eventually create the `ClusterCatalog` CRDs (I imagine fairly quickly after the informers are started) and the `Reflector` will eventually have a successful List/Watch operation.

If you would prefer we do something different, I'm open to exploring alternatives.

pkg/controller/dynamicrequiredmanifest_controller_test.go
```diff
@@ -161,3 +162,97 @@ func TestDynamicRequiredManifestControllerSync(t *testing.T) {
...
+					},
+				},
+			},
+			expectUpdate: true,
+			assertError:  noError(),
+		},
+		{
+			name:     "required and existing are deep derivative, no error, no update needed",
+			manifest: []byte(requiredYAML),
+			existing: &unstructured.Unstructured{
```
joelanford, line 213: Add a `metadata.labels` map with some custom values and set `Availability: Disabled` and make sure we still see "no updated needed"?

everettraven: https://github.com/openshift/cluster-olm-operator/pull/73/commits/b90b97c06be0d80e94126fd55109fa049daa75a1

pkg/clients/clients.go
```diff
@@ -154,6 +150,41 @@ func (ce ClusterExtensionClient) Informer() informers.GenericInformer {
 	return ce.informer
 }
 
+func NewClusterExtensionClient(dynClient dynamic.Interface) *ClusterExtensionClient {
+	infFact := dynamicinformer.NewDynamicSharedInformerFactory(dynClient, defaultResyncPeriod)
+	clusterExtensionGVR := ocv1alpha1.GroupVersion.WithResource("clusterextensions")
```
LalatenduMohanty, line 155: We did not have informer for `clusterextensions` before?

pkg/clients/clients.go
joelanford: Sounds good. Thanks for digging into it. I think we can leave as is, but we should keep this in mind to verify after this merges and https://github.com/openshift/operator-framework-catalogd/pull/81 is being tested.

pkg/controller/builder.go
```diff
@@ -33,10 +36,11 @@ type Builder struct {
 	ControllerContext *controllercmd.ControllerContext
 }
 
-func (b *Builder) BuildControllers(subDirectories ...string) (map[string]factory.Controller, map[string]factory.Controller, []configv1.ObjectReference, error) {
+func (b *Builder) BuildControllers(subDirectories ...string) (map[string]factory.Controller, map[string]factory.Controller, map[string]factory.Controller, []configv1.ObjectReference, error) {
```
LalatenduMohanty, line 39: We should add go doc for the method. Specially how we are starting new controller for each default catalog.

pkg/controller/dynamicrequiremanifest_controller.go
joelanford, line 1: Filename doesn't match the struct or test filename. Missing a `d` at the end of `required`.

pkg/controller/builder.go
```diff
@@ -102,10 +106,28 @@ func (b *Builder) BuildControllers(subDirectories ...string) (map[string]factory
 				return nil
 			}
 
+			if manifestGVK.Kind == "ClusterCatalog" && manifestGVK.Group == catalogdv1alpha1.GroupVersion.Group {
```
LalatenduMohanty, line 109: IMO we should add code comment around why creating controllers for each catalog is not an issue.

pkg/controller/dynamicrequiremanifest_controller.go
everettraven: Ah good catch!

PR #77, opened 2024-11-05
[HOLD] Add OLM resource to relatedObjects

perdasilva:


joelanford (approved):
/lgtm
/approve

cmd/cluster-olm-operator/main.go
```diff
@@ -189,3 +200,23 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 	<-ctx.Done()
 	return nil
 }
+
+// newOLMObjectReference creates a configv1.ObjectReference for
+// the cluster scoped OLM resources
+func newOLMObjectReference() configv1.ObjectReference {
+	return configv1.ObjectReference{
+		Group:    operatorv1alpha1.GroupName,
+		Resource: "OLM",
```
joelanford, line 209: /hold

Should this be `olms` to use the resource name, not the Kind?

manifests/0000_51_olm_07_cluster_operator.yaml
```diff
@@ -10,3 +10,10 @@ status:
   versions:
   - name: operator
     version: "0.0.1-snapshot"
+  relatedObjects:
+  - group: ""
+    name: openshift-cluster-olm-operator
+    resource: namespaces
+  - group: operator.openshift.io
+    name: cluster
+    resource: OLM
```
joelanford, line 19: Same question here about `olms` vs `OLM`.

perdasilva: this seems to work - though I agree, maybe the plurals would be better here. I'll update!

Thanks for calling this out!

everettraven (approved):
/lgtm

PR #82, opened 2024-11-13
NO-ISSUE: Remove library-go replaces and update clients.go for k8s 1.31

perdasilva:


grokspawn:
Supersedes https://github.com/openshift/cluster-olm-operator/pull/76

grokspawn:
/retitle  NO-ISSUE: Remove library-go replaces and update clients.go for k8s 1.31

PR #93, opened 2024-12-03
OCPBUGS-45098: Add infomer for proxy.config.openshift.io

tmshort:
Update the environment of the cluster-olm-operator
Update the env of the catalogd/operator-controller deployments

### DIFF
--- a/cmd/cluster-olm-operator/main.go
+++ b/cmd/cluster-olm-operator/main.go
@@ -24,6 +24,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/component-base/cli"
 	utilflag "k8s.io/component-base/cli/flag"
+	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-olm-operator/internal/utils"
 	"github.com/openshift/cluster-olm-operator/pkg/clients"
@@ -31,6 +32,10 @@ import (
 	"github.com/openshift/cluster-olm-operator/pkg/version"
 
 	catalogdv1 "github.com/operator-framework/catalogd/api/v1"
+)
+
+const (
+	olmProxyController = "OLMProxyController"
 )
 
 func main() {
@@ -149,6 +154,19 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		cc.EventRecorder.ForComponent("OLMIncompatibleOperatorController"),
 	)
 
+	// Side Effect! Update our environment immediately
+	err = controller.UpdateProxyEnvironment(klog.FromContext(ctx).WithName("main"), cl.ProxyClient)
+	if err != nil {
+		return err
+	}
+
+	proxyController := controller.NewProxyController(
+		olmProxyController,
+		cl.ProxyClient,
+		cl.OperatorClient,
+		cc.EventRecorder.ForComponent(olmProxyController),
+	)
+
 	versionGetter := status.NewVersionGetter()
 	versionGetter.SetVersion("operator", status.VersionForOperatorFromEnv())
 
@@ -174,7 +192,7 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 
 	cl.StartInformers(ctx)
 
-	for _, c := range append(staticResourceControllerList, upgradeableConditionController, incompatibleOperatorController, clusterOperatorController, operatorLoggingController) {
+	for _, c := range append(staticResourceControllerList, upgradeableConditionController, incompatibleOperatorController, clusterOperatorController, operatorLoggingController, proxyController) {
 		go func(c factory.Controller) {
 			defer runtime.HandleCrash()
 			c.Run(ctx, 1)
--- a/pkg/controller/builder.go
+++ b/pkg/controller/builder.go
@@ -2,6 +2,7 @@ package controller
 
 import (
 	"bytes"
+	"context"
 	"errors"
 	"fmt"
 	"io/fs"
@@ -18,6 +19,9 @@ import (
 	"github.com/openshift/library-go/pkg/operator/staticresourcecontroller"
 	"golang.org/x/text/cases"
 	"golang.org/x/text/language"
+
+	appsv1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -25,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/yaml"
+	"k8s.io/klog/v2"
 
 	"github.com/openshift/cluster-olm-operator/pkg/clients"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
@@ -102,13 +107,16 @@ func (b *Builder) BuildControllers(subDirectories ...string) (map[string]factory
 					b.Clients.OperatorClient,
 					b.Clients.KubeClient,
 					b.Clients.KubeInformerFactory.Apps().V1().Deployments(),
-					nil,
+					[]factory.Informer{
+						b.Clients.ProxyClient.Informer(),
+					},
 					[]deploymentcontroller.ManifestHookFunc{
 						replaceVerbosityHook("${LOG_VERBOSITY}"),
 						replaceImageHook("${CATALOGD_IMAGE}", "CATALOGD_IMAGE"),
 						replaceImageHook("${OPERATOR_CONTROLLER_IMAGE}", "OPERATOR_CONTROLLER_IMAGE"),
 						replaceImageHook("${KUBE_RBAC_PROXY_IMAGE}", "KUBE_RBAC_PROXY_IMAGE"),
 					},
+					UpdateDeploymentProxyHook(b.Clients.ProxyClient),
 				)
 				return nil
 			}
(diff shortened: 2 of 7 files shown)


tmshort:
/payload-job periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-proxy

tmshort:
/payload-job periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-proxy

tmshort:
> @tmshort: trigger 1 job(s) for the /payload-(with-prs|job|aggregate|job-with-prs|aggregate-with-prs) command
> 
> * periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-proxy
> 
> See details on https://pr-payload-tests.ci.openshift.org/runs/ci/3643cd20-b704-11ef-8e1c-d19e972aef3e-0

This job succeeded.

tmshort:
/retest

The **ci/prow/okd-scos-e2e-aws-ovn** job has been failing install, and generally has a high failure rate.

pkg/controller/proxycontroller.go
```diff
@@ -0,0 +1,74 @@
...
+	"github.com/go-logr/logr"
+	"github.com/openshift/library-go/pkg/controller/factory"
+	"github.com/openshift/library-go/pkg/operator/events"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/klog/v2"
+
+	"github.com/openshift/cluster-olm-operator/pkg/clients"
+)
+
+func NewProxyController(name string, proxyClient *clients.ProxyClient, operatorClient *clients.OperatorClient, eventRecorder events.Recorder) factory.Controller {
```
perdasilva, line 17: is it worth building out a unit test for this fella?

pkg/controller/proxycontroller.go
```diff
@@ -0,0 +1,74 @@
...
+			updateEnvironment(logger, "HTTP_PROXY", "")
+			updateEnvironment(logger, "HTTPS_PROXY", "")
+			updateEnvironment(logger, "NO_PROXY", "")
+			return nil
+		}
+		return fmt.Errorf("failed to get proxy: %w", err)
+	}
+
+	logger.Info("updating environment")
+	updateEnvironment(logger, "HTTP_PROXY", proxySpec.Status.HTTPProxy)
```
perdasilva, line 54: just a suggestion: maybe instead of taking the key/value as parameters, take a map? Then it might look like:

```
updateEnvironment(logger, map[string]string{
  "HTTP_PROXY"  : proxySpec.Status.HTTPProxy,
  "HTTPS_PROXY" : proxySpec.Status.HTTPSProxy,
  "NO_PROXY"    : proxySpec.Status.NO_PROXY,
})
```

cmd/cluster-olm-operator/main.go
```diff
@@ -149,6 +150,19 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
...
 	)
 
+	// Side Effect! Update our environment immediately
+	err = controller.UpdateProxyEnvironment(klog.FromContext(ctx).WithName("main"), cl.ProxyClient)
+	if err != nil {
+		return err
+	}
+
+	proxyController := controller.NewProxyController(
+		"OLMProxyController",
```
perdasilva, line 160: would it be worth moving all these controller ids to constants?

pkg/controller/builder.go
```diff
@@ -185,3 +191,38 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
 		return []byte(newDeployment), nil
 	}
 }
+
+func appendEnvIfNotPresent(env []corev1.EnvVar, name, value string) []corev1.EnvVar {
+	if value == "" {
+		return env
+	}
+	for _, e := range env {
+		if e.Name == name {
```
joelanford, line 200: If a proxy envvar is already present, this code will mean that we silently ignore the cluster's proxy configuration.

I kinda feel like we should either fail or overwrite.

Case for failing:
- The only reason the envvars would be present is if they show up in our manifests from operator controller or catalogd. We would probably want to know about the conflict of "use from manifest" vs "use from OCP environment" and decide how to handle it.

Case for overwriting:
- It seems extremely likely that if the fail case happens, we'll decide "we should use the cluster's proxy config". In that case, it seems like overwriting would make sense.

I lean toward failing so that we have a signal for this conflict and can make a decision about how to handle it at the time. I would rather deconflict in the static manifests vs runtime choices.

tmshort: What is the likelihood that we would want to _explicitly_ set these environment variables in the manifests? What is the use case? Is there a reason we _wouldn't_ want to use the cluster's proxy?

I would say it's a pretty low likelihood, and I don't really see any use cases, except maybe for testing?

You mention a "decision" between existing environment and new environment, but nothing that would be the deciding factor? (i.e. unless we just fail or overwrite, we'd need _something_ that indicates to use one or the other.) I don't think we want to add another parameter for that type of decision.

Honestly, I wasn't sure how to approach this. I note that there's no log here, so perhaps we should just overwti

cmd/cluster-olm-operator/main.go
tmshort: I'll do new ones, but I don't want to go and modify code outside the code of the proxy work.

pkg/controller/proxycontroller.go
tmshort: I can look, my concern would be in the function itself, as environments are not a simple `map`.

pkg/controller/proxycontroller.go
tmshort: There are admittedly few tests for the controllers, proper.

pkg/controller/builder.go
joelanford: My thinking is that if someone adds proxy env vars to upstream for some reason, those will make their way here. If this happens in 2 years and we forget that we have _this_ code from _this_ PR, we'll likely run into issues.

At that point, it might be nice for CI to fail on the downstream sync job with something like "detected conflicting envvars". The upstream change will be fresh on our minds at that point because the fact that it is in the sync PR means it is new from upstream.

tmshort:
/payload-job periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-proxy

pkg/controller/builder_test.go
```diff
@@ -1,8 +1,10 @@
-package controller
+package controller_test
```
joelanford, line 1: What necessitated this change? Seems like it has the undesired side-effect of needing to export some types and functions that don't otherwise need to be exported.

tmshort: Tests _should_ be a separate package... if we don't want to do that, then I can put it back, but yes, that was a side effect.

tmshort: (reverted)

tmshort:
/payload-job periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-proxy

For most recent code - the last time, the test process failed to execute?

tmshort:
I'm a little concerned that the proxy jobs are failing, and that the deployments are continuously being refreshed... large generation numbers and logs
Will have to investigate more, but this was fine before I implemented using the DeploymentHook...

pkg/controller/proxycontroller.go
tmshort: Seems as though using maps makes the application of the environment variables occur in a non-deterministic order, which was not happening before. I may need to get the keys and sort them.

tmshort: I found a reference for this behavior in "Learning Go"!

tmshort:
Now the map keys are sorted before being applied into the Deployment.

tmshort:
/payload-job periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-proxy

This should pass with the sorted map keys.

pkg/controller/builder.go
```diff
@@ -185,3 +195,60 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
...
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func UpdateDeploymentProxyHook(pc clients.ProxyClientInterface) deploymentcontroller.DeploymentHookFunc {
+	return func(_ *operatorv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		proxyConfig, err := pc.Get("cluster")
```
LalatenduMohanty, line 229: @tmshort I guess the proxy informer would provide us with the cluster wide proxy info. Is that right?

tmshort: Yes, and it's called `cluster`

cmd/cluster-olm-operator/main.go
```diff
@@ -149,6 +154,19 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		cc.EventRecorder.ForComponent("OLMIncompatibleOperatorController"),
 	)
 
+	// Side Effect! Update our environment immediately
+	err = controller.UpdateProxyEnvironment(klog.FromContext(ctx).WithName("main"), cl.ProxyClient)
```
joelanford, line 158: Should the logger use the "OLMProxyController" string as its name? `"main"` seems odd to me for the name of this thing.

```suggestion
	err = controller.UpdateProxyEnvironment(klog.FromContext(ctx).WithName(olmProxyController), cl.ProxyClient)
```

pkg/controller/builder.go
```diff
@@ -185,3 +195,60 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
 		return []byte(newDeployment), nil
 	}
 }
+
+func updateEnv(con *corev1.Container, name, value string) ([]corev1.EnvVar, error) {
```
joelanford, line 199: It feels a bit confusing to me that we accept a pointer to a container we want to update (which means we can update it in place), but then we return a new slice where the caller is expected to update the container, but also `append` has interesting semantics around whether the backing array is updated in place (if it has capacity) vs if a new backing array is created (when it doesn't)

WDYT about changing the function to do things the following way?
```go
func updateEnv(con *corev1.Container, name, value string) error {
   // check presence, and return err
   con.Env = append(con.Env, corev1.EnvVar{name, value})
   return nil
}
```

pkg/controller/builder.go
```diff
@@ -185,3 +195,60 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
...
+		}
+	}
+	if value == "" {
+		return con.Env, nil
+	}
+	klog.FromContext(context.Background()).WithName("builder").V(1).Info("Updated environment", "container", con.Name, "key", name, "value", value)
+	return append(con.Env, corev1.EnvVar{Name: name, Value: value}), nil
+}
+
+func setProxyEnvs(con *corev1.Container, vars map[string]string) error {
```
joelanford, line 212: Based on the function signature, I'd suggest a more generic name. Maybe `updateEnvFromMap`?

pkg/controller/builder.go
```diff
@@ -185,3 +195,60 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
...
+			if err != nil {
+				return err
+			}
+		}
+		for i := range deployment.Spec.Template.Spec.Containers {
+			err = setProxyEnvs(&deployment.Spec.Template.Spec.Containers[i], vars)
+			if err != nil {
+				return err
+			}
+		}
```
joelanford, line 251: Nit:
```go
var errs []error
// do for loops, collecting all the errors, but not returning
if len(errs) > 0 {
    return errors.Join(errs...)
}
return nil

File: pkg/controller/builder.go [hunk 11]
```diff
@@ -185,3 +195,60 @@ func replaceImageHook(placeholder string, desiredImageEnvVar string) deploymentc
...
+	klog.FromContext(context.Background()).WithName("builder").V(1).Info("Updated environment", "container", con.Name, "key", name, "value", value)
+	return append(con.Env, corev1.EnvVar{Name: name, Value: value}), nil
+}
+
+func setProxyEnvs(con *corev1.Container, vars map[string]string) error {
+	var err error
+	// The order these are put into the Container matter for hashing
+	// The ordering of maps varies by design in golang, so we have to sort a slice of keys
+	keys := maps.Keys(vars)
+	slices.Sort(keys)
```
joelanford, line 217: Complete nit, take it or leave it: An alternative to sorting the keys is to have something like this in the loop, and then not passing things through a map at all:
```go
for _, c := range containers {
  for _, envVar := range []corev1.EnvVar{
    {HTTPSProxy, proxyConfig.Status.HTTPSProxy},
    {HTTPProxy,  proxyConfig.Status.HTTPProxy},
    {NoProxy,    proxyConfig.Status.NoProxy},
  } {
    if err := updateEnv(c, envVar); err != nil {
      errs = append(errs, err)
    }
  }
}
```

pkg/controller/builder.go
tmshort: The original function was just taking in the env vars and modifying them and returning. I added the container to get the name, but didn't really consider the rest of the function...

pkg/controller/builder.go
tmshort: Yeah, that might be cleaner. @perdasilva suggested the use of a map (which caused some headaches), but this might be better, avoids the whole sorting thing and what-not...

tmshort:
/payload-job periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-proxy

pkg/controller/builder.go
LalatenduMohanty: +1 to fail CI when we find any conflicting envvars from upstream changes.

tmshort:
/payload-job periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-proxy

tmshort:
/payload-job periodic-ci-openshift-release-master-nightly-4.19-e2e-aws-ovn-proxy

Xia-Zhao-rh:
Test PASS, detail https://issues.redhat.com//browse/OCPBUGS-45098

/label qe-approved

joelanford (approved):

PR #96, opened 2024-12-11
NO-ISSUE: bump all bingo build dependencies

joelanford:
Seems like we need a golangci-lint bump in order to pass the lint test in #94.

Also, bumping bingo to 0.9 in order to ensure that the follow-up bump to go 1.23 goes okay (bingo 0.9 has a few bug fixes that are necessary when devs and CI are using go1.23)


LalatenduMohanty (approved):
/lgtm

PR #102, opened 2025-01-22
OPRUN-3663:  Watch and reconcile feature gates changes

azych:
Adds logic to watch and react to downstream feature gate changes by setting container env vars in `operator-controller-controller-manager` deployment manifests.
Incorporates the recommended `github.com/openshift/library-go` approach (see https://github.com/openshift/enhancements/blob/068e863988b58f70b5184e4ef49c0ad1c2913dfb/dev-guide/featuresets.md) to avoid duplicating this logic locally.

~As mentioned internally, there are still some things that need clarification - I'm raising them as comments here.~

@oceanc80 @joelanford

### DIFF
--- a/cmd/cluster-olm-operator/main.go
+++ b/cmd/cluster-olm-operator/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	goflag "flag"
 	"fmt"
 	"os"
@@ -191,6 +192,16 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 	operatorLoggingController := loglevel.NewClusterOperatorLoggingController(cl.OperatorClient, cc.EventRecorder.ForComponent("ClusterOLMOperatorLoggingController"))
 
 	cl.StartInformers(ctx)
+
+	log := klog.FromContext(ctx)
+	select {
+	case <-cl.FeatureGatesAccessor.InitialFeatureGatesObserved():
+		featureGates, _ := cl.FeatureGatesAccessor.CurrentFeatureGates()
+		log.Info("FeatureGates initialized", "knownFeatures", featureGates.KnownFeatures())
+	case <-time.After(1 * time.Minute):
+		log.Error(nil, "timed out waiting for FeatureGate detection")
+		return errors.New("timed out waiting for FeatureGate detection")
+	}
 
 	for _, c := range append(staticResourceControllerList, upgradeableConditionController, incompatibleOperatorController, clusterOperatorController, operatorLoggingController, proxyController) {
 		go func(c factory.Controller) {
--- a/internal/featuregates/mapper.go
+++ b/internal/featuregates/mapper.go
@@ -1,0 +1,105 @@
+package featuregates
+
+import (
+	"bytes"
+	"errors"
+	"strings"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
+)
+
+// Add your new upstream feature gate here
+// const (
+// 		MyUpstreamFeature = "MyUpstreamFeature"
+// )
+
+type MapperInterface interface {
+	OperatorControllerUpstreamForDownstream(downstreamGate configv1.FeatureGateName) []string
+	OperatorControllerDownstreamFeatureGates() []configv1.FeatureGateName
+	CatalogdUpstreamForDownstream(downstreamGate configv1.FeatureGateName) []string
+	CatalogdDownstreamFeatureGates() []configv1.FeatureGateName
+}
+
+// Mapper knows the mapping between downstream and upstream feature gates for both OLM components
+type Mapper struct {
+	operatorControllerGates map[configv1.FeatureGateName][]string
+	catalogdGates           map[configv1.FeatureGateName][]string
+}
+
+func NewMapper() *Mapper {
+	// Add your downstream to upstream mapping here
+	operatorControllerGates := map[configv1.FeatureGateName][]string{
+		// features.FeatureGateNewOLMMyDownstreamFeature: {MyUpstreamControllerOperatorFeature}
+	}
+	catalogdGates := map[configv1.FeatureGateName][]string{
+		// features.FeatureGateNewOLMMyDownstreamFeature: {MyUpstreamCatalogdFeature}
+	}
+
+	for _, m := range []map[configv1.FeatureGateName][]string{operatorControllerGates, catalogdGates} {
+		for downstreamGate := range m {
+			// features.FeatureGateNewOLM is a GA-enabled downstream feature gate.
+			// If there is a need to enable upstream alpha/beta features in the downstream GA release
+			// get approval via a merged openshift/enhancement describing the need, then carve out
+			// an exception in this failsafe code
+			if downstreamGate == features.FeatureGateNewOLM {
+				panic(errors.New("FeatureGateNewOLM used in mappings"))
+			}
+			if !strings.HasPrefix(string(downstreamGate), string(features.FeatureGateNewOLM)) {
+				panic(errors.New("all downstream feature gates must use NewOLM prefix by convention"))
+			}
+		}
+	}
+
+	return &Mapper{operatorControllerGates: operatorControllerGates, catalogdGates: catalogdGates}
+}
+
+// OperatorControllerDownstreamFeatureGates returns a list of all downstream feature gates
+// which have an upstream mapping configured for the operator-controller component
+func (m *Mapper) OperatorControllerDownstreamFeatureGates() []configv1.FeatureGateName {
+	return getKeys(m.operatorControllerGates)
+}
+
+// CatalogdDownstreamFeatureGates returns a list of all downstream feature gates
+// which have an upstream mapping configured for the catalogd component
+func (m *Mapper) CatalogdDownstreamFeatureGates() []configv1.FeatureGateName {
+	return getKeys(m.catalogdGates)
+}
+
+// OperatorControllerUpstreamForDownstream returns upstream feature gates which are configured
+// for a given downstream feature gate for the operator-controller component
+func (m *Mapper) OperatorControllerUpstreamForDownstream(downstreamGate configv1.FeatureGateName) []string {
+	return m.operatorControllerGates[downstreamGate]
+}
+
+// CatalogdUpstreamForDownstream returns upstream feature gates which are configured
+// for a given downstream feature gate for the catalogd component
+func (m *Mapper) CatalogdUpstreamForDownstream(downstreamGate configv1.FeatureGateName) []string {
+	return m.catalogdGates[downstreamGate]
+}
+
+// FormatAsEnabledArgs combines list of feature gate names into
+// an all-enabled arg format of <feature_gate_name1>=true,<feature_gate_name1>=true etc.
+func FormatAsEnabledArgs(enabledFeatureGates []string) string {
+	buf := bytes.Buffer{}
+	for _, gateName := range enabledFeatureGates {
+		buf.WriteString(gateName)
+		buf.WriteRune('=')
+		buf.WriteString("true")
+		buf.WriteRune(',')
+	}
+	if buf.Len() > 0 {
+		// get rid of trailing ','
+		buf.Truncate(buf.Len() - 1)
+	}
+
+	return buf.String()
+}
+
+func getKeys(m map[configv1.FeatureGateName][]string) []configv1.FeatureGateName {
+	keys := make([]configv1.FeatureGateName, 0, len(m))
+	for k := range m {
+		keys = append(keys, k)
+	}
+	return keys
+}
--- a/pkg/clients/clients.go
+++ b/pkg/clients/clients.go
@@ -21,11 +21,16 @@ import (
 	operatorv1apply "github.com/openshift/client-go/operator/applyconfigurations/operator/v1"
 	operatorclient "github.com/openshift/client-go/operator/clientset/versioned"
 	operatorinformers "github.com/openshift/client-go/operator/informers/externalversions"
+	internalfeatures "github.com/openshift/cluster-olm-operator/internal/featuregates"
 	"github.com/openshift/library-go/pkg/apiserver/jsonpatch"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
+	"github.com/openshift/library-go/pkg/operator/configobserver/featuregates"
+	"github.com/openshift/library-go/pkg/operator/events"
 	"github.com/openshift/library-go/pkg/operator/resource/resourceapply"
+	"github.com/openshift/library-go/pkg/operator/status"
 	"github.com/openshift/library-go/pkg/operator/v1helpers"
 	ocv1 "github.com/operator-framework/operator-controller/api/v1"
+	corev1 "k8s.io/api/core/v1"
 	apiextensionsclient "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -62,6 +67,8 @@ type Clients struct {
 	KubeInformerFactory        informers.SharedInformerFactory
 	ConfigInformerFactory      configinformer.SharedInformerFactory
 	KubeInformersForNamespaces v1helpers.KubeInformersForNamespaces
+	FeatureGatesAccessor       featuregates.FeatureGateAccess
+	FeatureGateMapper          internalfeatures.MapperInterface
 }
 
 func New(cc *controllercmd.ControllerContext) (*Clients, error) {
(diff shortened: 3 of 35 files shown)


pkg/controller/builder.go
```diff
@@ -250,3 +253,38 @@ func UpdateDeploymentProxyHook(pc clients.ProxyClientInterface) deploymentcontro
...
+		klog.FromContext(context.Background()).WithName("builder").V(0).Info("FeatureGatesHook updating environment", "deployment", deployment.Name)
+
+		featuregates, err := fa.CurrentFeatureGates()
+		if err != nil {
+			return fmt.Errorf("error getting featuregates.config.openshift.io/cluster: %w", err)
+		}
+
+		vars := []corev1.EnvVar{{Name: "FEATURE_GATES"}}
+		if featuregates.Enabled(features.FeatureGateNewOLM) {
+			vars[0].Value = string(features.FeatureGateNewOLM)
```
azych, line 268: I wasn't sure to which upstream feature gates we want to have this mapped to, so I'm currently just setting this to the name of the our only one downstream feature gate.
Once the mapping is known, we can adjust and probably lowercase the value(s) here as well.

Additional point - doing the mapping here will introduce fragility with upstream feature gate names, but I am not sure there is a way around it. Deferring it to upstream and pulling an additional openshift dependency isn't probably what we want.

pkg/controller/builder.go
```diff
@@ -250,3 +253,38 @@ func UpdateDeploymentProxyHook(pc clients.ProxyClientInterface) deploymentcontro
...
+		}
+
+		var errs []error
+		for i := range deployment.Spec.Template.Spec.InitContainers {
+			err = setContainerEnv(&deployment.Spec.Template.Spec.InitContainers[i], vars)
```
azych, line 273: As mentioned in the description, this logic is currently using the same approach as with proxy config - set appropriate env vars in containers. If we wanted to move feature gates to args, I think we'd have to add handling to upstream first.

cmd/cluster-olm-operator/main.go
```diff
@@ -86,6 +90,26 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		return err
 	}
 
+	operatorNamespace := "openshift-cluster-olm-operator"
```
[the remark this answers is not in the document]
joelanford: +1 for env using downward API.

pkg/clients/clients.go
```diff
@@ -126,6 +128,9 @@ func New(cc *controllercmd.ControllerContext) (*Clients, error) {
 }
 
 func (c *Clients) StartInformers(ctx context.Context) {
+	if c.FeatureGatesAccessor != nil {
+		go c.FeatureGatesAccessor.Run(ctx)
+	}
```
joelanford, line 133: If `c.FeatureGatesAccessor` is `nil` here, that's a programmer error right? If so, seems reasonable to let a nil pointer error panic happen like we would with any of the other struct fields?

pkg/controller/builder.go
```diff
@@ -219,34 +222,56 @@ func setContainerEnv(con *corev1.Container, envs []corev1.EnvVar) error {
...
+		if err != nil {
+			return fmt.Errorf("error getting featuregates.config.openshift.io/cluster: %w", err)
 		}
-		for i := range deployment.Spec.Template.Spec.Containers {
-			err = setContainerEnv(&deployment.Spec.Template.Spec.Containers[i], vars)
-			if err != nil {
-				errs = append(errs, err)
-			}
+
+		vars := []corev1.EnvVar{{Name: "FEATURE_GATES"}}
```
joelanford, line 249: Does the upstream feature gate code know about the `FEATURE_GATES` env? I have always used the `--feature-gates` flag upstream.

pkg/controller/builder.go
joelanford: It might be good to go ahead and setup a package and/or type that facilitates registering the mapping from downstream feature gates to upstream feature gates. That would help others have a very clear pattern to follow when new downstream feature gates come along.

> doing the mapping here will introduce fragility with upstream feature gate names

I agree there is some fragility here, but downstream seems like the appropriate place to handle the mapping. I would hope that we wouldn't have to deal with this all that often, other than adding new mappings.

cmd/cluster-olm-operator/main.go
```diff
@@ -86,6 +90,26 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
...
+		APIVersion: "apps/v1",
+		Kind:       "Deployment",
+		Namespace:  operatorNamespace,
+		Name:       "cluster-olm-operator",
+	})
+
+	operatorImageVersion := status.VersionForOperatorFromEnv()
+	desiredVersion := operatorImageVersion
+	missingVersion := "0.0.1-snapshot"
+	// By default, this will exit(0) the process if the featuregates ever change to a different set of values.
```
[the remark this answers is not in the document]
joelanford: Yep, I think it is better not to exit and restart if we don't have to. If/when we have a feature gate that directly affects cluster-olm-operator itself, maybe we reassess. (And it might be worth putting that extra bit of detail in the comment)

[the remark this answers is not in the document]
azych: sounds good, I'll include that in the comment

cmd/cluster-olm-operator/main.go
```diff
@@ -86,6 +90,27 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
...
+	missingVersion := "0.0.1-snapshot"
+	featureGateAccessor := featuregates.NewFeatureGateAccess(
+		desiredVersion,
+		missingVersion,
+		cl.ConfigInformerFactory.Config().V1().ClusterVersions(), cl.ConfigInformerFactory.Config().V1().FeatureGates(),
+		eventRecorder,
+	)
+	// modify the default behavior of calling exit(0) to noop
+	featureGateAccessor.SetChangeHandler(func(featureChange featuregates.FeatureChange) {})
+	cl.FeatureGatesAccessor = featureGateAccessor
```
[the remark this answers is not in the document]
azych: yes, I think we should be able to hide this in there together with other ~client setup

pkg/controller/builder.go
azych: that's a good question.
Just on a face value, if `--feature-gates` arg is supported, afaik most of the libraries should be able to handle env var form as well, but I will have to check that upstream to be sure.
If it won't or there won't be too much work required to change this to an arg, I'll do that since it seems it's usually passed this way.

cmd/cluster-olm-operator/main.go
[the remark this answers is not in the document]
azych: I used the OperatorNamespace from clustercontext we pass to clients. It gets the namespace from within the cluster or an env if set, so should be good

pkg/controller/builder.go
azych: I switched to args and mapped to only one valid upstream feature gate which seems to be accepted (even though AFAIR it has been recently removed from upstream upstream ;)) and was able to successfully validate the behavior together with manager actually reading the feature gates that are set

azych:
@joelanford please take a look, it should be good to go with the testing I've done. Regarding the one remaining comment and current upstream feature gates situation - should we wait for the new catalogd api feature gate? is using the one I mentioned here https://github.com/openshift/cluster-olm-operator/pull/102#discussion_r1928522765 - ok?
Thanks

internal/featuregates/mapper.go
```diff
@@ -0,0 +1,60 @@
...
+	"bytes"
+	"fmt"
+
+	configv1 "github.com/openshift/api/config/v1"
+	"github.com/openshift/api/features"
+)
+
+const (
+	// Add your new upstream gate here
+	ForceSemverUpgradeConstraintsGate = "ForceSemverUpgradeConstraints"
```
joelanford, line 13: I think I'd leave the actual code completely empty of upstream/downstream feature gates since:
- the only downstream feature gate we have right now is enabled in GA (so any upstream features it maps to would also be suddenly enabled in GA, which we don't want)
- we don't actually want to enable this particular feature gate at all (As you mentioned, I fully removed it upstream earlier this week)

What I would suggest is some comments in the various places with obvious dummy values like
```
// MyUpstreamFeature = "MyUpstreamFeature"
...
// features.FeatureGateNewOLMDownstreamFeature: {MyUpstreamFeature},
```

internal/featuregates/mapper.go
```diff
@@ -0,0 +1,60 @@
...
+type Mapper struct {
+	gates map[configv1.FeatureGateName][]string
+}
+
+func NewMapper() *Mapper {
+	// Add your downstream to upstream mapping here
+	gates := map[configv1.FeatureGateName][]string{
+		features.FeatureGateNewOLM: {ForceSemverUpgradeConstraintsGate},
+	}
+
```
joelanford, line 30: For now, I think we should loop through the configured gates and `panic` if:
- the gate name is `NewOLM` (we don't want to accidentally turn a non-default-enabled feature gate on in production.
- they don't have a `NewOLM` prefix (we want all sub features of OLMv1 to be "under" `NewOLM` which is the parent feature gate for _all_ of OLMv1.

We should probably explain this in the panic message, and then maybe include a comment that suggests that relaxing these rules may be necessary if we ever decide we want to enable an alpha or beta feature from upstream in OCP GA downstream, or if a completely separate OCP feature impacts how we would deploy our components.

internal/featuregates/mapper.go
```diff
@@ -0,0 +1,60 @@
...
+	// Add your new upstream gate here
+	ForceSemverUpgradeConstraintsGate = "ForceSemverUpgradeConstraints"
+)
+
+type MapperInterface interface {
+	UpstreamForDownstream(downstreamGate configv1.FeatureGateName) ([]string, error)
+}
+
+// Mapper knows the mapping between downstream and upstream feature gates
+type Mapper struct {
```
joelanford, line 21: I suspect that somewhere in here, we'll need to distinguish between catalogd feature gates and operator-controller feature gates.

Maybe the mapping should be that a downstream feature gate maps to two slices of upstream feature gates. One for catalogd and one for operator-controller.

joelanford: Another potentially annoying nuance to consider: What if a single component (say catalogd) in the future gets another deployment or a new container in an existing deployment's pod spec. We won't be able to blindly apply feature gate flags to every single container. Commands that don't recognize the flag or the feature gate will fail.

On the other hand: maintaining a specific mapping of "component name -> deployment name -> container name -> feature gate set" seems onerous and brittle.

We don't have this problem right now, so maybe we punt. But:
- Todd and I were recently discussing adding init containers in order to configure multiple certificates robustly
- We _just_ removed kube-rbac-proxy as a sibling container from catalogd and operator-controller

So point being, these are very real possibilities in the short-term.

pkg/controller/builder.go
```diff
@@ -247,6 +252,62 @@ func UpdateDeploymentProxyHook(pc clients.ProxyClientInterface) deploymentcontro
...
+}
+
+func UpdateDeploymentFeatureGatesHook(
+	featuresAccessor featuregates.FeatureGateAccess,
+	featuresMapper internalfeatures.MapperInterface,
+) deploymentcontroller.DeploymentHookFunc {
+	return func(_ *operatorv1.OperatorSpec, deployment *appsv1.Deployment) error {
+		klog.FromContext(context.Background()).WithName("builder").V(0).Info("FeatureGatesHook updating environment", "deployment", deployment.Name)
+
+		featuregates, err := featuresAccessor.CurrentFeatureGates()
```
joelanford, line 267: Is it possible to iterate `featuregates`? My hope is that this hook code would not have to change whenever we add/remove downstream feature gates.

I was thinking something like:
```
var upstreamGates []string
for _, fg := range featuregates {
    // featuresMapper.UpstreamForDownstream(fg) returns nil if we
    // don't have a mapping for the downstream feature (we'll see _all_
    // of the OCP feature gates, but we only care about ours)
    upstreamGates = append(upstreamGates, featuresMapper.UpstreamForDownstream(fg))
}
```

pkg/controller/builder.go
```diff
@@ -247,6 +252,62 @@ func UpdateDeploymentProxyHook(pc clients.ProxyClientInterface) deploymentcontro
...
+			}
+			argVal = internalfeatures.FormatAsEnabledArgs(upstreamGates)
+		}
+
+		var errs []error
+		for i := range deployment.Spec.Template.Spec.Containers {
+			klog.FromContext(context.Background()).WithName("builder").V(4).Info("FeatureGatesHook iterating containers", "container", deployment.Spec.Template.Spec.Containers[i].Name)
+			if !strings.EqualFold(deployment.Spec.Template.Spec.Containers[i].Name, "manager") {
+				continue
+			}
```
joelanford, line 286: This is saying "only add the feature gates flag if the container name is manager"? If so, +1. I think that is a reasonable convention. And it's one of the answers I had in mind for my earlier question about dealing with the question of which containers to inject this flag into.

joelanford:
> should we wait for the new catalogd api feature gate?

I don't think we should wait. Let's put together what we think is a good starting point. When @anik120 and @grokspawn get to the point where they need to add their mapping, we can always tweak/fix things we missed in this first pass.

pkg/controller/builder.go
azych: that's a good idea and I think we should be able to leverage `Mapper` to provide a list of downstream features that are of interest here. Mapper has that information already (and it was meant to have it) and this way we should be able to avoid having to loop through all features.

internal/featuregates/mapper.go
azych: I've added panic checks and comments. I thought about just leaving the `NewOLM` as an empty mapping with a comment, but better to have an additional verification for now at least and it costs us nothing.

Regarding the 'NewOLM' prefix - I've added panic check as well, but if it ever triggers it would mean that a non-prefixed flag has already passed review and has been added to `github.com/openshift/api`, thinking out loud here, it would be good to be able to place a failsafe like that directly in the api.

internal/featuregates/mapper.go
azych: I've split the mappings between `catalogd` and `operator-controller` and also added hook logic to make sure we're honoring the distinguishment.
I could not rely only on a container name (https://github.com/openshift/cluster-olm-operator/pull/102#discussion_r1929135234) alone as both deployments use 'manager' and had to include deployment name as well.

And ofc I'll be interested to be included in those discussions if possible.

pkg/controller/builder.go
azych: yes, that was the idea here, but with the both components having separate sets of feature flags, I had to include deployment name in the mix, see https://github.com/openshift/cluster-olm-operator/pull/102#discussion_r1930344434

internal/featuregates/mapper.go
joelanford: There are a bunch of other features in openshift/api that are completely unrelated to `NewOLM` and OLMv1. So my thinking here is that this is our way of saying "any OLMv1-specific features _MUST_ be prefixed with `NewOLM`".

I agree, it would be better to have that failsafe in `openshift/api`, but I'm not sure how realistic that is given all the other features for other OCP components that are registered there.

pkg/controller/builder.go
ankitathomas: Can we address some of the fragility by moving the operator-controller/catalogd features package out of internal and importing it directly to figure out all supported upstream feature gates?

azych:
> > should we wait for the new catalogd api feature gate?
> 
> I don't think we should wait. Let's put together what we think is a good starting point. When @anik120 and @grokspawn get to the point where they need to add their mapping, we can always tweak/fix things we missed in this first pass.

@joelanford I think it should be good to go™ in the current state, WDYT?

internal/featuregates/mapper.go
```diff
@@ -0,0 +1,106 @@
...
+		// features.FeatureGateNewOLMMyDownstreamFeature: {MyUpstreamControllerOperatorFeature}
+	}
+	catalogdGates := map[configv1.FeatureGateName][]string{
+		// features.FeatureGateNewOLMMyDownstreamFeature: {MyUpstreamCatalogdFeature}
+	}
+
+	for _, m := range []map[configv1.FeatureGateName][]string{controllerGates, catalogdGates} {
+		for downstreamGates, upstreamGates := range m {
+			// features.FeatureGateNewOLM is a GA-enabled by default feature gate
+			// only remove this fail-safe if you want to add by-default enabled upstream feature(s)
```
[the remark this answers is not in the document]
azych: Interesting, I thought the mapping here would be our source of truth when it comes to which upstream features are enabled. So I imagined that our current GA downstream feature flag (NewOLM) would have a mapping here to the 'default' set of upstream flags that we want to enable and then other downstream sub-feature gates would follow the same pattern. 
If you're saying there's a way for upstream to override/circumvent this then I fully agree about it not being an ideal situation.

internal/featuregates/mapper.go
```diff
@@ -0,0 +1,106 @@
...
+			if downstreamGates == features.FeatureGateNewOLM {
+				panic(errors.New("FeatureGateNewOLM used in mappings"))
+			}
+			if slices.ContainsFunc(upstreamGates, func(s string) bool {
+				return !strings.HasPrefix(s, string(features.FeatureGateNewOLM))
+			}) {
+				panic(errors.New("all upstream feature gates must use NewOLM prefix"))
+			}
```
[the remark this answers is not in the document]
azych: good catch

azych:
@joelanford @anik120 @oceanc80 can you approve and label the changes so they can be merged? Last pass from @joelanford was all nits/simple renames.

okd-scos-e2e-aws-ovn [passed before recent rename/comment changes](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_cluster-olm-operator/102/pull-ci-openshift-cluster-olm-operator-main-okd-scos-e2e-aws-ovn/1884129663542890496) and I will not be re-running them. They were failing before that with the same code that they passed with and are optional.

grokspawn:
/retitle OPRUN-3663:  Watch and reconcile feature gates changes

PR #105, opened 2025-02-20
NO-ISSUE: Bump kubernetes and openshift/api dependencies

anik120:
Bump openshift/api to make https://github.com/openshift/api/pull/2202/files#diff-a8b6135d50534471326ea7bcd20e0f5eae25353f7788338060f718128a6a0b34R525 available

$ go get github.com/openshift/api@744790f2cff777b1bb29bdccce10e4e84cff0a69 $ go mod tidy
$ go mod vendor

Also needed
$ go get openshift/client-go@f7ec47e

to resolve vendoring inconsistency.


go.mod
```diff
@@ -21,13 +19,13 @@ require (
...
 	golang.org/x/text v0.19.0
 	helm.sh/helm/v3 v3.16.2
-	k8s.io/api v0.31.2
+	k8s.io/api v0.32.2
 	k8s.io/apiextensions-apiserver v0.31.2
-	k8s.io/apimachinery v0.31.2
-	k8s.io/client-go v0.31.2
+	k8s.io/apimachinery v0.32.2
+	k8s.io/client-go v0.32.2
 	k8s.io/component-base v0.31.2
```
grokspawn, line 26: bump?

anik120: yea fixing it I need a moment, this is automatically being set to v0.31.2 I probably have to use replace directive.

go.mod
```diff
@@ -1,15 +1,13 @@
 module github.com/openshift/cluster-olm-operator
 
-go 1.22.5
-
-toolchain go1.22.7
+go 1.23.4
```
joelanford, line 3: openshift/api looks like this:

https://github.com/openshift/api/blob/744790f2cff777b1bb29bdccce10e4e84cff0a69/go.mod#L3-L5

Do we actually need `go 1.23.4` or can we set it to `go 1.23.0`?

go.mod
joelanford: > this is automatically being set to v0.31.2 I probably have to use replace directive.

That doesn't seem right 🤔

go.mod
```diff
@@ -21,13 +19,13 @@ require (
...
 	k8s.io/apiextensions-apiserver v0.31.2
-	k8s.io/apimachinery v0.31.2
-	k8s.io/client-go v0.31.2
+	k8s.io/apimachinery v0.32.2
+	k8s.io/client-go v0.32.2
 	k8s.io/component-base v0.31.2
 	k8s.io/klog/v2 v2.130.1
-	k8s.io/utils v0.0.0-20240921022957-49e7df575cb6
+	k8s.io/utils v0.0.0-20241104100929-3ea5e8cea738
 	sigs.k8s.io/controller-runtime v0.19.0
```
joelanford, line 29: Bump controller-runtime to v0.20.2?

go.mod
anik120: Didn't end up needing the replace, sorry I was in the middle of brainstorming session that time 😅

cmd/cluster-olm-operator/main.go
```diff
@@ -75,6 +75,7 @@ func newStartCommand() *cobra.Command {
 		"cluster-olm-operator",
 		version.Get(),
 		runOperator,
+		nil,
```
anik120, line 78: Does this look right? 

The documentation for clock.Clock [here](https://github.com/openshift/library-go/blob/master/pkg/controller/controllercmd/cmd.go#L81) reads 

```
// Clock allows for injecting fake or real clocks into code that
// needs to do arbitrary things based on time.
```

cmd/cluster-olm-operator/main.go
```diff
@@ -187,6 +188,7 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		cl.OperatorClient,
 		versionGetter,
 		cc.EventRecorder.ForComponent("olm"),
+		nil,
```
anik120, line 191: And here

cmd/cluster-olm-operator/main.go
tmshort: `nil` is probably find here and below.

cmd/cluster-olm-operator/main.go
grokspawn, line 191:
```suggestion
		cc.Clock,
```

cmd/cluster-olm-operator/main.go
grokspawn: I did 
```golang
clock.RealClock{}
```
but I was literally cargo-culting another example which did nothing with the resulting clock.

joelanford: It is useful to look through the change that introduced this new function parameter to understand what is expected. https://github.com/openshift/library-go/pull/1880

TL;DR (but I really would encourage reviewing the details in this case and others when things like this arise): we need to pass a non-nil clock. `clock.RealClock{}` seems like the correct choice in `main.go`

Makefile
```diff
@@ -3,6 +3,8 @@ SHELL := /usr/bin/env bash
 GO_TEST_PACKAGES :=./pkg/... ./cmd/...
 GO_BUILD_BINDIR := bin
 
+export GOTOOLCHAIN=go1.23.4
+
```
joelanford, line 7: Can we remove this from the PR? I feel like figuring out what we want to do with GOTOOLCHAIN is deserving of its own PR and discussion.

cmd/cluster-olm-operator/main.go
grokspawn: Yeah, that's [where](https://github.com/openshift/library-go/pull/1880/files#diff-8a3d8d95d4d7548a8fcf8e71fd086374dd3098542f86d7440f41ed42eb974222R38) I cargo-culted it from since I figured it would do no harm but wasn't 100% on why it was desirable. 😅

Makefile
anik120: This was the first thing that this PR needed because any vendor update (ie any modification to the go.mod file) was being accompanied by an automatic go version update 😄 

I'm trying to think if there's a good follow up to capture that'll make us come back to this? Maybe an issue that says "let's revisit this"?

joelanford: That's because of your environment, not anything intrisically related to bumping openshift/api and kubernetes repos.

cmd/cluster-olm-operator/main.go
anik120: Mea culpa, between the go version sneakily updating behind the scenes and all the slack messages, I was totally trying to rush this in unnecessarily 😅 let me see if I can find some explanation to confirm `clock.RealClock{}` is the right answer here.

Makefile
anik120: So even for my environment, when I was downloading and installing go1.23.4 binary explicitly, when I checked go version it was automatically downloading and installing go1.23.5 by replacing go1.23.4 (which would then update the gotoolchain in the go.mod file). We talked about it in this [slack thread](https://redhat-internal.slack.com/archives/C080C4VDMDZ/p1740064903531859?thread_ts=1739996798.737199&cid=C080C4VDMDZ) too and finally came to this conclusion. 

However, looks like I've managed to pin the go version to 1.23.4 locally for myself. Let me remove this change and see if I can avoid the unexpected go version patch update.

anik120: I just looked again and you're right, we most likely don't need this change in the Makefile (go.mod file already has the go version pinned), removing it..

cmd/cluster-olm-operator/main.go
```diff
@@ -187,6 +189,7 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 		cl.OperatorClient,
 		versionGetter,
 		cc.EventRecorder.ForComponent("olm"),
+		cc.Clock,
```
grokspawn, line 192: Still have reservations about passing in a nil PassiveClock here.

pkg/clients/clients.go
```diff
@@ -172,6 +172,7 @@ func setupFeatureGatesAccessor(
 			Namespace:  operatorNamespace,
 			Name:       "cluster-olm-operator",
 		},
+		nil,
```
grokspawn, line 175: Still have reservations about supplying a nil PassiveClock here.

grokspawn: https://github.com/openshift/cluster-olm-operator/blob/4a5b8c1578f4befccdef4cd21bd99a480a1400f8/pkg/clients/clients.go#L109C1-L110C1 establishes a clock which can be passed into this method by widening the signature and passing in `opClient.Clock` to this function.

anik120: Oops yea I missed this, thanks for catching it. I just passed in `clock.RealClock{}` since we're using this no-op clock anyway, and that doesn't feel like it's enough justification to widen the function signature.

cmd/cluster-olm-operator/main.go
anik120: Update: `clock.RealClock{}` is needed otherwise a nil pointer error is thrown.

go.mod
anik120: I changed this to look identical to what we have in openshift/api 👍🏽

pkg/clients/clients.go
grokspawn: I mean ... it's not an exported interface.

tmshort:
Looks reasonable to me.

pkg/clients/clients.go
anik120: agreed, but it still feels like "effort for nothing", no?

.bingo/golangci-lint.mod
```diff
@@ -1,5 +1,7 @@
 module _ // Auto generated by https://github.com/bwplotka/bingo. DO NOT EDIT
 
-go 1.22.1
+go 1.23.0
 
-require github.com/golangci/golangci-lint v1.62.2 // cmd/golangci-lint
+toolchain go1.23.4
```
joelanford, line 5: This toolchain line is not present in the go.mod for golangci-lint 1.64.5. I suspect this is another artifact of your environment. Can you remove this?

https://github.com/golangci/golangci-lint/blob/0a603e49e5e9870f5f9f2035bcbe42cd9620a9d5/go.mod#L3

pkg/clients/clients.go
grokspawn: I'd almost always prefer to perpetuate an object instance rather than get in the habit of needlessly generating new instances.  
So... no.

anik120:
Clicked the wrong button by mistake 😮‍💨

PR #109, opened 2025-03-24
OCPBUGS-48478: Add openshift node selector annotation

oceanc80:
Add openshift node selector annotation


kuiwang02:
/hold

for pre-merge testing

thanks

kuiwang02:
/unhold
/label qe-approved

but the OCPBUGS-54179 is not resolved and the upgrade still fails. it seems to fix the openshift-catalogd ns and openshift-operator-controller ns as well.

PR #128, opened 2025-07-09
OPRUN-3999: Validate feature-gate arguments

tmshort:
This changes the behavior of the feature gate argument settings into the deployments.

Originally, an error would occur if there were any feature-gate aguments present. This was removed in the first pass of the feature gated API changes.

Instead, the old feature-gate arguments were removed, and then were replaced by any new arguments.

Now, this checks to ensure that any feature-gate arguments from the experimental manifest match what we are setting. If there is a mismatch, it means that there is a mismatch in what we think should be enabled feature gates in cluster-olm-operator and in operator-framework-operator-controller.

These sets of feature-gates are expected to match.

This should impact only techpreview builds, which will break until both are are set the same.

However, a new flag: -warn-on-arg-mismatch will avoid the error allowing the the deployment to proceed. This flag is added to the cluster-olm-operator deployment, and should be removed when we branch (at which point, we want to error if the args don't match).


tmshort:
I'm sorry... we errored out in _deprovision_, and that caused the whole test to fail‽‽
/retest-required

internal/featuregates/mapper_test.go
```diff
@@ -58,8 +58,8 @@ func TestFormatAsEnabledArgs(t *testing.T) {
 		},
 		{
 			name:     "multiple feature gates",
-			in:       []string{"testGate1", "testGate2", "testGate3"},
-			expected: "testGate1=true,testGate2=true,testGate2=true",
```
joelanford, line 62: How did this test pass before? Input was 1,2,3. Expected output was 1,2,2?

pkg/controller/featuregates_hook_test.go
```diff
@@ -257,9 +257,10 @@ func TestUpdateDeploymentFeatureGatesHook(t *testing.T) {
 		}
 		mockMapper.ValidateCalls(t, 1, 1, 0, 0)
 
+		// The new code will now validate that the arguments are the same, and accept that as valid
 		err = update(nil, dep)
-		if err == nil {
-			t.Fatal("no error in second update")
+		if err != nil {
+			t.Fatal("error in second update")
 		}
```
joelanford, line 264: Seeing a bunch of `err == nil` flip to `err != nil` is making my spidey sense go off.

The test scenarios seemed to originally assume that an error _would_ occur. Now, because of the code changes, the scenarios mean that an error does not occur. To me, it feels like the appropriate thing to do is change the test inputs so that the test assertions continue to hold, and then add some new tests to cover the various success/error outcomes we expect from the new code.

internal/featuregates/mapper_test.go
tmshort: I don't believe it was run!

tmshort: I added `./internal` into `GO_TEST_PACKAGES` variable in the Makefile

pkg/controller/builder.go
```diff
@@ -253,23 +260,61 @@ func UpdateDeploymentProxyHook(pc clients.ProxyClientInterface) deploymentcontro
 	}
 }
 
-func setContainerArg(con *corev1.Container, arg, value string) {
+// The return behavior:
+// 1. If arg type does not already exist, and we're only adding arg = success (i.e. adding only case)
```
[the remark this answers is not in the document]
tmshort: Upstream and downstream have different sets of feature gates enabled.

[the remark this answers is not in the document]
tmshort: IMHO, the feature gates and manifest must match; this is because there may be resources missing or extra that may not work. So, perhaps a stricter results may be necessary.

pkg/controller/featuregates_hook_test.go
tmshort: Yeah, running the update a second time is now a no-op, so perhaps it makes sense to remove it.

internal/featuregates/mapper_test.go
tmshort: And apparently, we don't have a unit-test CI‽‽

joelanford: 🙃

pkg/controller/builder.go
```diff
@@ -253,23 +254,61 @@ func UpdateDeploymentProxyHook(pc clients.ProxyClientInterface) deploymentcontro
...
+	}
+
+	// Check to see if the container values match the new values
+	if value != containerValue {
+		// Arguments do not match, so return an error.
+		// The likely reason for this error is that the feature flag coding has not completed yet;
+		// either the cluster-olm-operator mapping is not present,
+		// or the feature was not added to the experimental manifest.
+		err := fmt.Errorf("new and existing arguments do not match")
+		log.Error(err, "New and existing arguments do not match", "arg", arg, "new", value, "existing", containerValue)
```
joelanford, line 298: This log line will look something like this:
```
"ERROR: New and existing arguments do not match", "err": "new and existing arguments do not match", "arg": "foo", "new": "bar1", "existing": "bar2"
```

Note the duplicate error text. Do we need this log line (or any log line in this function)? Now that there is no warning possibility, it seems like either everything is good or we return an error, that will eventually be logged somewhere.

pkg/controller/builder.go
```diff
@@ -253,23 +254,61 @@ func UpdateDeploymentProxyHook(pc clients.ProxyClientInterface) deploymentcontro
...
+
+	// Update the args, this will make a previously multiline argument into a single line.
 	if value != "" {
 		log.Info("Updated args", "container", con.Name, "arg", arg, "value", value)
 		args = append(args, fmt.Sprintf("%s=%s", arg, value))
 	}
+
+	// Update the containers arguments
 	con.Args = args
+	return nil
```
joelanford, line 313: It seems like this function could be simplified quite a bit by:
1. Loop through the container args
2. For each arg, use `strings.Cut(arg, "=")` to split into `name` and `value`
3. If the `name` doesn't match the passed in `arg`, just `continue`
4. Use `strings.Join(slices.Sorted(slices.Values(strings.Split(val, ","))), ",")` to construct comparable existing and desired values
5. If they don't match return the error; if they do, note that we found the arg
6. After the loop, if we have not found the arg, add it to con.Args

joelanford: Ah, now I see the nuance of handling multiple `--feature-gates` arguments, and needing to collect their values. I do still wonder if we could simplify it some though.

joelanford: Another couple of thoughts:
- the function name and the actual function implementation seem to have drifted
- is this functionality still actually applicable to any arbitrary argument that might show up in the container args? i feel like it is different enough where we may want to specifically look for `--feature-gates`  args and handle those separately.

tmshort: But yes, this is much more specific for `--feature-flags", and we don't process any other flags. And due to specialized behavior, yeah, we might want to rename it.

tmshort: I am not finding `List` in `sets`... (i.e. https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets)
```
	haveValues := strings.Join(sets.List(foundValues), ",")
```
Unfortunately there's `UnsortedList()` (but not `List()` or `SortedList()`, so it still requires sorting. And `slices.Sorted()` wants some type of iterator, so I can't use the output of `UnsortedList()` directly...

But I think I got it...

pkg/controller/builder.go
tmshort: Ywah, I knew it would be duplicate test. I removed all the logs; they were only useful for debugging.

pkg/controller/builder.go
joelanford: For future reference, `set.List` is here: https://github.com/kubernetes/apimachinery/blob/v0.33.3/pkg/util/sets/set.go#L201

But what you have now is effectively the same, and is symmetric with computing `wantValues`, so looks good to me.

joelanford (approved):

PR #132, opened 2025-07-22
OCPBUGS-58992: Bump helm to v3.18.4

anik120:
Also note that helmv3.18.4 requires gov1.24.0, so
gov1.23.0 has been updated to gov1.24.0 as a result.


go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/openshift/cluster-olm-operator
 
-go 1.23.0
+go 1.24.0
 
-toolchain go1.23.4
+toolchain go1.24.5
```
tmshort, line 5: OCP builds don't support go1.24.5, it supports go1.24.4.

grokspawn: helm just requires [1.24.0](https://github.com/helm/helm/blob/98d6d9c1efeb7707dd331484645c48e1215baf05/go.mod#L3), so we should be able to do the same here, no?

go.mod
```diff
@@ -15,35 +15,34 @@ require (
...
+	github.com/spf13/pflag v1.0.6
+	github.com/stretchr/testify v1.10.0
+	golang.org/x/net v0.40.0
+	golang.org/x/text v0.26.0
+	helm.sh/helm/v3 v3.18.4
+	k8s.io/api v0.33.2
+	k8s.io/apiextensions-apiserver v0.33.2
+	k8s.io/apimachinery v0.33.2
+	k8s.io/client-go v0.33.2
+	k8s.io/component-base v0.33.2
```
camilamacedo86, line 28: Are those the identical versions used in upstream?

anik120: We don't have an upstream cluster-olm operator, and this looks like it's aligned with https://github.com/openshift/operator-framework-operator-controller/blob/main/go.mod#L36-L40 so I think we're good here.

camilamacedo86 (approved):
/lgtm

go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/openshift/cluster-olm-operator
 
-go 1.23.0
+go 1.24.0
 
-toolchain go1.23.4
+toolchain go1.24.4
```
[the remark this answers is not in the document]
grokspawn: IMO, this is the proper approach.  
the toplevel directive communicates that we require some version of 1.24.  
the toolchain directive communicates that some dependency requires a higher version.

PR #134, opened 2025-07-25
OPRUN-4034: Handle CustomNoUpgrade better

tmshort:
If we are running CustomNoUpgrade, ignore the flag mismatch, and just set the flags. No error is generated.

### DIFF
--- a/pkg/controller/featuregates_hook.go
+++ b/pkg/controller/featuregates_hook.go
@@ -27,6 +27,7 @@ const (
 func UpdateDeploymentFeatureGatesHook(
 	featuresAccessor featuregates.FeatureGateAccess,
 	featuresMapper internalfeatures.MapperInterface,
+	featureSet configv1.FeatureSet,
 ) deploymentcontroller.DeploymentHookFunc {
 	return func(_ *operatorv1.OperatorSpec, deployment *appsv1.Deployment) error {
 		logger := klog.FromContext(context.Background()).WithName("feature_gates_hook")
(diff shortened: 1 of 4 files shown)


pkg/controller/featuregates_hook_test.go
```diff
@@ -487,4 +487,67 @@ func TestUpdateDeploymentFeatureGatesHook(t *testing.T) {
...
+			t.Fatalf("args length not 1: %+v", dep)
+		}
+		expectedArg := "--feature-gates=TestUpstreamGate1=true,TestUpstreamGate2=true"
+		if expectedArg != dep.Spec.Template.Spec.Containers[0].Args[0] {
+			t.Fatalf("args differ, container: %q, expected: %q", dep.Spec.Template.Spec.Containers[0].Args[0], expectedArg)
+		}
+		mockMapper.ValidateCalls(t, 0, 0, 2, 1)
+	})
+
+	t.Run("catalog mapping exists with mismatch feature gates and matching container and DevPreviewNoUpgrade", func(t *testing.T) {
```
grokspawn, line 527:
```suggestion
	t.Run("catalogd mapping exists with mismatch feature gates and matching container and DevPreviewNoUpgrade", func(t *testing.T) {
```

pkg/controller/featuregates_hook.go
```diff
@@ -64,7 +65,10 @@ func UpdateDeploymentFeatureGatesHook(
 			if !strings.EqualFold(deployment.Spec.Template.Spec.Containers[i].Name, "manager") {
 				continue
 			}
-			if err := setContainerFeatureGateArg(&deployment.Spec.Template.Spec.Containers[i], featureGatesToSet); err != nil {
+
+			// If featureSet is not found in AllFixedFeatureSets, then it Custom, so we need to ignore mismatches
```
grokspawn, line 69:
```suggestion
			// If featureSet is not found in AllFixedFeatureSets, then it is Custom, so we need to ignore mismatches
```
?
Is there any such thing as a "custom feature set", or are you just covering all the bases?

PR #139, opened 2025-09-17
OCPBUGS-61873: Update RBAC to update/patch/delete what we create

tmshort:
If we can create it, we can patch, update.

If we can update it, we can patch (and vice-versa)

Reduce our requirements on clusteroperators

We also need to be able to bind to cluster-admin for future boxcutter support

### DIFF
--- a/manifests/0000_51_olm_02_operator_clusterrole.yaml
+++ b/manifests/0000_51_olm_02_operator_clusterrole.yaml
@@ -40,6 +40,8 @@ rules:
       - list
       - watch
       - create
+      - update
+      - patch
   - apiGroups:
       - ""
     resources:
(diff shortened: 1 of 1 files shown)


tmshort:
/payload-job periodic-ci-openshift-release-master-ci-4.21-upgrade-from-stable-4.20-e2e-aws-ovn-upgrade

manifests/0000_51_olm_02_operator_clusterrole.yaml
```diff
@@ -40,6 +40,9 @@ rules:
       - list
       - watch
       - create
+      - update
+      - patch
+      - delete
```
wking, line 45: > If we can create it, we can patch, update, delete
>
> Except events, clusteroperators: we can't delete

I can see "if we can `create`, we can `update` and `patch`".  I'm less clear on `delete`.  Maybe you can put that off until you have operator code-paths you can point to that actually attempt to delete resources?  Because `create`, the update verbs, and the view verbs give you enough to manage a forever-child, and you only need `delete` if you decide you no longer need that child resource.  And global ConfigMap `delete` (and similar for the other `resources`) seems like a pretty broad power, that is worth doing without unless you actually need that power.

tmshort: That's fair, we will need delete at some point, but it can be limited in scope.

tmshort:
/payload-job periodic-ci-openshift-release-master-ci-4.21-upgrade-from-stable-4.20-e2e-aws-ovn-upgrade

joelanford:
/lgtm

/hold
So we can verify the result of the upgrade test.

PR #184, opened 2026-03-27
OPRUN-4521: handle OCP 4.23/5.0 upgrade boundary

fgiudici:
OCP 4.23 and OCP 5.0 are considered equivalent releases, so 4.23 clusters cannot upgrade to OCP 5.0.
Next allowed release is 5.1.
Normalize currentOCPMinorVersion 4.23 → 5.0 before comparing with operator's maxOpenShiftVersion, tighten regex to reject leading-zero minor versions, and extend test coverage for the boundary cases.

[OPRUN-4521](https://redhat.atlassian.net/browse/OPRUN-4521)


tmshort (commented):
Generally fine.

I just prefer the terms "upgradable"/"not upgradable" over "compatible"/"incompatible", because this is really about whether the cluster can be upgraded, so using "upgradable" makes more sense to me.

oceanc80 (commented):
/lgtm

PR #199, opened 2026-05-01
[release-4.19] OCPBUGS-80420: Bump google.golang.org/grpc to v1.79.3

ocp-sustaining-admins:
This is an automated PR from the ARC system.

**Go version used:** `go1.23.12`

**Patch applied:** `github.com/openshift-sustaining/grpc-go@v1.75.1-sec.1`

**Modules affected:** `.`

The commands used to generate this PR were:

```
go mod edit -replace google.golang.org/grpc=github.com/openshift-sustaining/grpc-go@v1.75.1-sec.1
go mod tidy
go mod vendor
```

A member of the Red Hat Openshift Sustaining Team will review the PR and take appropriate action.

## Summary by CodeRabbit

* **Chores**
  * Updated Go dependencies to latest versions for improved stability and security.
  * Optimized continuous integration linter configuration to reduce resource consumption during builds.
  * Updated internal import declarations to use standard library components.


MrSanketkumar:
@coderabbitai review

vendor/modules.txt
```diff
@@ -1480,3 +1493,4 @@ sigs.k8s.io/structured-merge-diff/v4/value
 ## explicit; go 1.12
 sigs.k8s.io/yaml
 sigs.k8s.io/yaml/goyaml.v2
+# google.golang.org/grpc => github.com/openshift-sustaining/grpc-go v1.75.1-sec.1
```
tmshort, line 1496: This is v1.75.1, whereas the fix is looking for v1.79.3; does this contain the proper fix?

MrSanketkumar: Yes, it includes the proper fix from v1.79.3. Since we have forked it, we created multiple patched versions for different Go versions to ensure compatibility and avoid breaking changes.

That’s why we are using a `replace` directive along with this patched version, which is specifically built to remain compatible with this OCP version, as it currently uses Go 1.23.

tmshort:
/lgtm
/approve

I looked at the go.mod/go.sum files. The replacement line ought to contain the patch; as the recommended fix requires upgrading to a more recent version of golang.

PR #203, opened 2026-05-05
OPRUN-4599: dynamically resolve catalog image tag from OCP release version

tmshort:
openshift.yaml may set options.openshift.catalogs.version to the
sentinel value "ocp-release" to opt into dynamic resolution. At
startup, compute the catalog tag from RELEASE_VERSION and substitute
it wherever the sentinel appears in the Helm values before rendering.

## Summary by CodeRabbit

* **New Features**
  * Operator reads its image/version once and derives a cluster catalog image tag from the OCP minor version.
  * Helm rendering can be conditionally overridden to use the computed catalog tag when applicable; Helm values can now be inspected via dot-delimited keys.

* **Tests**
  * Added unit tests validating catalog image tag formatting (vMAJOR.MINOR) and the Helm override/values behavior.

PR #210, opened 2026-05-29
OPRUN-4627: override catalog tag to 4.x when RELEASE_VERSION is 4.x and catalogd.yaml pins v5.0

tmshort:
Partially reverts #203: removes the ocp-release sentinel mechanism (GetCatalogImageTag, ClusterCatalogImageTag, applyCatalogImageTagOverride, and the associated Builder field and main.go changes).

Replaces it with a targeted check in renderHelmTemplate: when openshift/helm/catalogd.yaml (operator-framework-operator-controller) pins options.openshift.catalogs.version to "v5.0" and RELEASE_VERSION is a 4.x stream (the 4.23/5.0 co-release), substitute the 4.x catalog tag so images match the release stream. For 5.x releases, including 5.1 running against a "v5.0" or "v5.1" catalog, the value in catalogd.yaml is left untouched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Catalog image tag resolution: OCP 5.1 is special-cased to use the v5.0 catalog tag; 4.x releases rewrite a loaded v5.0 catalog tag to the matching v<major>.<minor>, while 5.x releases preserve the loaded tag.

* **Refactor**
  * Controller setup simplified to compute and apply catalog tags at render time instead of precomputing/storing them.

* **Tests**
  * Legacy table test removed and replaced with focused tests driving behavior by release version.

### DIFF
--- a/internal/versionutils/version_utils.go
+++ b/internal/versionutils/version_utils.go
@@ -61,12 +61,6 @@ func ToAllowedSemver(data []byte) (*semver.Version, error) {
 	return &version, nil
 }
 
-// GetCatalogImageTag converts an OCP version to the catalog image tag format used by default catalogs.
-// For example, version 4.22.0 returns "v4.22", and version 5.0.0 returns "v5.0".
-func GetCatalogImageTag(version *semver.Version) string {
-	return fmt.Sprintf("v%d.%d", version.Major, version.Minor)
-}
-
 // ocpVersion500 is the semver representation of OCP 5.0, which is co-released with 4.23 as an
 // equivalent release. Neither upgrades to the other; both upgrade exclusively to 5.1.
 var ocpVersion500 = semver.Version{Major: 5, Minor: 0, Patch: 0}
--- a/cmd/cluster-olm-operator/main.go
+++ b/cmd/cluster-olm-operator/main.go
@@ -230,13 +230,6 @@ func runOperator(ctx context.Context, cc *controllercmd.ControllerContext) error
 
 	log := klog.FromContext(ctx)
 
-	operatorImageVersion := status.VersionForOperatorFromEnv()
-	currentOCPMinorVersion, err := versionutils.GetCurrentOCPMinorVersion(operatorImageVersion)
-	if err != nil {
-		return err
-	}
-	catalogImageTag := versionutils.GetCatalogImageTag(currentOCPMinorVersion)
-
 	fg, err := cl.ConfigClient.ConfigV1().FeatureGates().Get(ctx, "cluster", metav1.GetOptions{})
 	if err != nil {
 		return fmt.Errorf("unable to retrieve featureSet: %w", err)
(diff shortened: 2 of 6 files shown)


tmshort:
Related to: https://github.com/openshift/operator-framework-operator-controller/pull/734

internal/versionutils/version_utils.go
```diff
@@ -64,6 +64,12 @@ func ToAllowedSemver(data []byte) (*semver.Version, error) {
 // GetCatalogImageTag converts an OCP version to the catalog image tag format used by default catalogs.
 // For example, version 4.22.0 returns "v4.22", and version 5.0.0 returns "v5.0".
 func GetCatalogImageTag(version *semver.Version) string {
+	// TODO: Remove after 4.23/5.0 branch cut; on the main branch replace "ocp-release" in
+	// openshift/helm/catalogd.yaml (github.com/openshift/operator-framework-operator-controller) with "v5.0".
+	// OCP 5.1 uses v5.0 catalog images until the 5.1 catalogs are available.
+	if version.Major == 5 && version.Minor == 1 {
+		return "v5.0"
+	}
 	return fmt.Sprintf("v%d.%d", version.Major, version.Minor)
```
joelanford: Something that was brought up today by @grokspawn. Instead of using `ocp-release` as a sentinel for "auto-version", it might be better to have specific code that handles _just_ the "map 5.0 to 4.23" case:

```go
name, tag := extractNameTag(imageRef)
if tag == "v5.0" and currentClusterVersion == "4.23" {
  tag = "v4.23"
}
imageRef = makeImageRef(name, tag)
```

If we did that, 5.1 would use 5.0 automatically until we change it, just like we've always done.

tmshort: I hear you. It's still code that needs to be deleted; one we branch, and are past 4.23 and no longer need this code.

The `ocp-release` sentinel could be reused; which is why I implemented it as generic. 

Unfortunately, it doesn't seem to be a solution that isn't temporary and becomes tech-debt almost immediately and then needs to be deleted. I'm really disliking all solutions right now (says the person who doesn't want to deal with temporary, special-case code).

joelanford: Yeah, I think it's tech debt no matter what. But it seems like even more tech debt to have code that handles versions outside of 5.0 and 4.23. Seems like the approach Jordan and I were mulling over minimizes the tech debt and keeps our normal processes otherwise undisturbed?

tmshort: I basically think I did as suggested.

tmshort:
/override ci/prow/openshift-e2e-aws-customnoupgrade

This is unable to install, through no fault of OLM

ankitathomas:
/verified by @ankitathomas 

- Verified catalog versions do not change when `options.openshift.catalogs.version`  is not `v5.0` regardless of RELEASE_VERSION
- Verified catalog versions change to match RELEASE_VERSION with major version 4 when `options.openshift.catalogs.version` is `v5.0`
- Verified catalog versions do not change to match RELEASE_VERSION with major version other than 4 when `options.openshift.catalogs.version` is `v5.0`

PR #211, opened 2026-06-05
OPRUN-4640: Update to golang 1.26.3 and openshift-4.23 builders

rashmigottipati:


tmshort:
The `openshift-e2e-aws-techpreview` failure was `sig-network`, so if it continues to flake, we can override it.

tmshort:
/lgtm

This is small enough that I can approve and lgtm

PR #214, opened 2026-06-15
OCPBUGS-88736: Apply cluster TLS security profile to the metrics server

tmshort:
The cluster-olm-operator metrics server (port 8443) was using default TLS settings rather than the cluster-wide TLS security profile from apiservers.config.openshift.io/cluster. The profile was already applied to operator-controller and catalogd via UpdateDeploymentObservedConfigHook, but not to the operator's own serving endpoint.

Fix this with two coordinated mechanisms:

1. Pre-configure metrics server TLS at startup: a PersistentPreRunE hook in newStartCommand reads the cluster TLS profile from the APIServer before controllercmd starts the metrics server, writes a GenericOperatorConfig temp file with the correct minTLSVersion and cipherSuites, and sets --config to point to it. controllercmd's SetRecommendedHTTPServingInfoDefaults preserves non-empty values, so the TLS profile flows into the metrics server. The hook is skipped if --config is already explicitly provided.

2. Restart on profile change: runOperator watches the operator client informer (the OLM CR) rather than the APIServer directly, so the handler fires only after TLSObserverController has already written the new settings to observedConfig and the deployment hook has had a chance to update the operand Deployments. When the stored TLS profile diverges from what was read at startup, the handler overwrites the config file with the new TLS settings. controllercmd's existing WithRestartOnChange mechanism detects the file content change and performs a graceful restart, after which PersistentPreRunE re-reads the new profile and the metrics server starts with correct TLS.

## Summary by CodeRabbit

* **New Features**
  * The operator’s metrics server now monitors the cluster’s TLS security profile and automatically updates its TLS configuration.
  * When the minimum TLS version or cipher suites change, the metrics server gracefully restarts to apply the new settings.
* **Tests**
  * Added unit coverage for TLS profile extraction, config generation, and metrics server TLS serving info behavior, including not-found and error scenarios.

### DIFF
--- a/cmd/cluster-olm-operator/main.go
+++ b/cmd/cluster-olm-operator/main.go
@@ -6,12 +6,15 @@ import (
 	goflag "flag"
 	"fmt"
 	"os"
+	"reflect"
 	"time"
 
 	configv1 "github.com/openshift/api/config/v1"
 	"github.com/openshift/cluster-olm-operator/internal/versionutils"
 
 	_ "github.com/openshift/api/operator/v1/zz_generated.crd-manifests"
+	configclient "github.com/openshift/client-go/config/clientset/versioned"
+	libgoclient "github.com/openshift/library-go/pkg/config/client"
 	"github.com/openshift/library-go/pkg/controller/controllercmd"
 	"github.com/openshift/library-go/pkg/controller/factory"
 	"github.com/openshift/library-go/pkg/operator/loglevel"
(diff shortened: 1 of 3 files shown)


tmshort:
/retest

Failures not OLM-related.

rashmigottipati (approved):

sg00dwin:
**Verified on a 5.0 cluster (5.0.0-0.ci-2026-06-18-055849)**

Used Claude Code to build the PR binary from branch HEAD (0e4b771f), package it in a container, and deploy to the cluster with a CVO override. TLS probing was done with `gnutls-cli` and `openssl s_client` via `oc port-forward` to port 8443.

**Bug confirmed on stock image:** With profile set to Modern, metrics port still accepted TLS 1.2.

**Results with PR image:**

| Profile | TLS 1.0 | TLS 1.1 | TLS 1.2 | TLS 1.3 | Expected | Result |
|---------|---------|---------|---------|---------|----------|--------|
| Default (Intermediate) | NO | NO | YES | YES | 1.2+ | PASS |
| Modern | NO | NO | NO | YES | 1.3 only | PASS |
| Intermediate | NO | NO | YES | YES | 1.2+ | PASS |
| Old | YES | YES | YES | YES | 1.0+ | PASS |

Pod correctly restarted on each profile change (3 restarts across 3 transitions), confirming the restart-on-config-change mechanism works.

PR #216, opened 2026-07-16
OPRUN-4696: fix upgrade block message and add 5.0 cluster unit tests

tmshort:
Update the incompatible operator message from "4.23 or 5.0" to "5.1": OCP 4.23 and 5.0 are co-released equivalents whose only upgrade target is 5.1, so the message should name that target.

Add missing controller-level unit tests for the 5.0 cluster side of the 4.23/5.0 boundary: maxOCPVersion=5.0 and maxOCPVersion=4.23 both block on a 5.0 cluster (next target is 5.1). Also add a test asserting the message contains "5.1" to lock in the fix.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved incompatibility messaging by dynamically determining the next upgrade target version, including correct handling for 4.23 → 5.1 and subsequent 5.0/5.1 upgrade paths.
  * Expanded detection and messaging validation for incompatible operators during 5.0 cluster upgrades, including both Boxcutter and Helm-managed operators.

* **Tests**
  * Added unit tests to verify next OCP minor version calculations and updated controller test coverage for additional 5.0-related scenarios.

### DIFF
--- a/pkg/controller/incompatible_operator_controller.go
+++ b/pkg/controller/incompatible_operator_controller.go
@@ -92,8 +92,7 @@ func (c *incompatibleOperatorController) sync(ctx context.Context, _ factory.Syn
 		// deterministic ordering
 		sort.Strings(incompatibleOperators)
 
-		// TODO(4.23): Update this message when main becomes 4.23 development
-		message := fmt.Sprintf("Found ClusterExtensions that require upgrades prior to upgrading cluster to version 4.23 or 5.0: %s.", strings.Join(incompatibleOperators, ","))
+		message := fmt.Sprintf("Found ClusterExtensions that require upgrades prior to upgrading cluster to version %s: %s.", versionutils.NextOCPMinorVersion(*c.currentOCPMinorVersion), strings.Join(incompatibleOperators, ","))
 		if err != nil {
 			message += fmt.Sprintf("\n Additionally the following errors were encountered while getting extension metadata: %s", err.Error())
 		}
--- a/pkg/controller/incompatible_operator_controller_test.go
+++ b/pkg/controller/incompatible_operator_controller_test.go
@@ -423,6 +423,133 @@ func TestIncompatibleOperatorController_Sync(t *testing.T) {
 				conditionReason: "",
 			},
 		},
+		// 5.0 cluster cases: maxOCPVersion=5.0 and maxOCPVersion=4.23 both block (next target is 5.1).
+		{
+			name: "boxcutter: maxOCPVersion=5.0 with cluster at 5.0 is not upgradable",
+			args: args{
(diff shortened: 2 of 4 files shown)


pkg/controller/incompatible_operator_controller.go
```diff
@@ -92,8 +92,7 @@ func (c *incompatibleOperatorController) sync(ctx context.Context, _ factory.Syn
 		// deterministic ordering
 		sort.Strings(incompatibleOperators)
 
-		// TODO(4.23): Update this message when main becomes 4.23 development
-		message := fmt.Sprintf("Found ClusterExtensions that require upgrades prior to upgrading cluster to version 4.23 or 5.0: %s.", strings.Join(incompatibleOperators, ","))
+		message := fmt.Sprintf("Found ClusterExtensions that require upgrades prior to upgrading cluster to version 5.1: %s.", strings.Join(incompatibleOperators, ","))
```
joelanford: Do we want to hardcode `5.1` here? I would think we have have code like:

```go
nextVersion := fmt.Sprintf("%d.%d", curMajor, curMinor + 1)
if curVersion == "4.23" {
   nextVersion := "5.1"
}
return nextVersion
```

pkg/controller/incompatible_operator_controller_test.go
```diff
@@ -423,6 +423,97 @@ func TestIncompatibleOperatorController_Sync(t *testing.T) {
...
+				clusterObjectSets: []runtime.Object{
+					createObjectSet("test-operator-rev1", 1, "Active", "test-operator", "test-bundle-1.0", olmPropertyAnnotation(`[{"type":"olm.maxOpenShiftVersion","value":"5.0"}]`)),
+				},
+				currentOCPVersion: "5.0.0",
+				boxCutterEnabled:  true,
+			},
+			wants: wants{
+				conditionStatus:   operatorv1.ConditionFalse,
+				conditionReason:   reasonIncompatibleOperatorsInstalled,
+				messageContaining: "test-bundle-1.0",
```
joelanford, line 442: Might want to test the message containing the next expected minor version as well?

I know the 4.22 ship has sailed, but in theory we can setup the code such that it fully handles the `4.22 -> [4.23,5.0] -> 5.1` upgrade variants now that we know all of the possibilities.

But we should at least set the code and tests up to handle:
- 4.23 -> 5.1
- 5.y -> 5.y+1

So we won't have to revisit this again until the 6.0 cutover.

fgiudici (commented):
/lgtm

tmshort:
The failed CI devpreview/techpreview is sig-instrumentation complaining about pods not having secure metric endpoints; and they aren't _our_ pods. So, I overrode them.

PR #219, opened 2026-07-21
adjust RBAC rolenames to adjust for serviceacct deprecation

grokspawn:
upstream changes to deprecate clusterextension.serviceaccount have RBAC changes which will not work with the current set of manifests.  This PR proposes changes to respect clusters which establish either the new or old RBAC instead. 

Enables https://github.com/openshift/operator-framework-operator-controller/pull/767

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved operator startup compatibility by recognizing the current cluster-admin role binding in addition to legacy binding names.
  * Helps clean up leftover legacy role bindings after RBAC is ready, supporting smoother upgrades across versions.

### DIFF
--- a/cmd/cluster-olm-operator/main.go
+++ b/cmd/cluster-olm-operator/main.go
@@ -52,6 +52,7 @@ const (
 	// operator-controller resource identifiers
 	operatorControllerNamespace      = "openshift-operator-controller"
 	operatorControllerServiceAccount = "operator-controller-controller-manager"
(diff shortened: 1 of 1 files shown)


tmshort:
The renaming of resources could lead to orphaning of the older resources, something to potentially watch for, COO doesn't have a mechanism to clean up those old resources on an upgrade.

tmshort:
This should be tested with the sync PR (or vice-versa).

pedjak:
let's merge https://github.com/openshift/cluster-olm-operator/pull/222 instead.

PR #223, opened 2026-07-22
OCPBUGS-95475: chore: bump containerd to 1.7.33

grokspawn:
## Summary by CodeRabbit

* **Chores**
  * Updated an underlying container runtime dependency to a newer version, improving maintenance and compatibility.


tmshort:
/verified by CI

I'm less concerned about the devpreview/techpreview

tmshort:
I had overrriden techpreview/devpreview in another PR because sig-instrumentation was complaining about pods (that aren't OLMv1) didn't have secured metric endpoints. I'd be ok with doing the same here, if they fail the same way.

